PR #618, opened 2020-06-04
Change/multiple data types

abkfenris:
For both the current conditions graphs and table, it matched by individual data types. This meant that platforms that have multiple sensor packages reading the same values, there might be multiple sets of the same information displayed.

Now you can specify a list of data types to display, and it find the first matching one to display.

CU-7rdn5m

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@ Additions:
 
 Changes:
 
-- Remove signal flags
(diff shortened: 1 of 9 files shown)


abkfenris:
Task linked: [CU-7rdn5m Preferred and ordered data types for current conditions](https://app.clickup.com/t/7rdn5m)

PR #639, opened 2020-06-17
Limit the current conditions table to a 15 min period

abkfenris:
CU-44dw8h[in review]


abkfenris:
Task linked: [CU-44dw8h Hide from current conditions table when reading is old](https://app.clickup.com/t/44dw8h)

Issue #118, opened 2019-05-23
Dealing with missing data in charts


youngmorse:
In the observations tab charts, missing values being plotted as zeroes. Seeing this currently in Buoy E water temperature and sig wave height. Can see in ERDDAP that values are not being reported for those parameters. Need to treat missing values as null if possible for highcharts.

<img width="1392" alt="Screen Shot 2019-05-23 at 1 41 37 PM" src="https://user-images.githubusercontent.com/3279645/58274279-87315380-7d60-11e9-8f48-78f6d507fece.png">

<img width="1391" alt="image" src="https://user-images.githubusercontent.com/3279645/58274386-c3fd4a80-7d60-11e9-8670-9be412034cac.png">

PR #880, opened 2020-11-03
Filter current conditions table by date and strip nulls on the client side

abkfenris:
Stripping nulls on the client side reduces the number of constraints needed (each constraint combo adds an extra request when there are multiple datasets to display).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ Changes:
 
 Fixes:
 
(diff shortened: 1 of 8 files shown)


abkfenris:
Closed  #118

abkfenris:
Task linked: [CU-bundr4 Missing data values are plotting as 0 in charts](https://app.clickup.com/t/bundr4)

PR #973, opened 2021-01-04
Update URI encoding method for ERDDAP 2.x

abkfenris:


abkfenris:
Cypress tests pass with `docker-compose run -e REACT_APP_PROXY_ROOT=http://localhost:9090/ -p 3000:3000 client yarn start` There currently is an issue with Cypress disagreeing with the CORS proxy. If using the same proxy locally, or with port-forwards there isn't an issue, so it may be the Kubernetes Ingress or at some other point.

PR #979, opened 2021-01-06
Fix app manifest to work with iOS

### LINKED ISSUE #977: Saving as iOS shortcut results in blank screen
When a user saves a specific page to their Home Screen, they instead just get a page with only the header and footer. 

This looks to be because the `src/public/manifest.json` defined a [`start_url`](https://www.w3.org/TR/appmanifest/#start_url-member) that does not contain a valid route. It looks like removing the `start_url` from the manifest may do the trick by allowing the current route to be persisted instead. 

Reported by @mattboutet

abkfenris:
The `manifest.json` specified a `start_url` to a non existent route. This caused the user's selected page to be overridden.

Also brought in the latest favorite icons.
simulator_screenshot_F4019883-AE03-4460-A16B-03375B8325C9
simulator_screenshot_B9991640-A5F5-4149-89A0-2F58BB0AB80E

Thanks to @mattboutet for reporting!

Closes #977


abkfenris:
Tests pass with `docker-compose run -e REACT_APP_PROXY_ROOT=http://localhost:9090/ -p 3000:3000 client yarn start` There currently is an issue with Cypress disagreeing with the CORS proxy. If using the same proxy locally, or with port-forwards there isn't an issue, so it may be the Kubernetes Ingress or at some other point.

Issue #1216, opened 2021-04-16
'Compare' Page


abkfenris:
Right now we're have a nice display of recent and forecast data for single platforms. What we don't currently have is a route to compare data from different platforms.

I think we could add __Compare__ as top level navigation (NERACOOS, Home, Regions, Compare, About). This way it is clear that it requires extra interaction compared to the existing views, and doesn't complicate them.

![image](https://user-images.githubusercontent.com/1296209/115038413-84343480-9e9d-11eb-8bd4-4ab99898c412.png)

User flow:

- Users would start by selecting a measure/data type they want to view (Sea Water Temperature, Barometric Pressure, Wind Speed..., maybe default to Wind Speed?)
- The map will display the the platforms with the selected data type
  - The latest value will be displayed
  - The color of the dot behind the value will vary based on the range of values
    - [Colorbrewer sequential color map?](https://colorbrewer2.org/#type=sequential&scheme=OrRd&n=3)
  - Platforms can be tapped on to select them
  - Selected platforms are circled with a border
  - If a platform has data at multiple depths, the depth closest to the surface will be displayed
  - Other platforms will be grayed out
- Below the measure selection, the timeseries for the data type can be selected (in addition to being able to select them from the map)
  - If multiple depths are available from a platform, they are separately select able here
- Data Type and time series selection drives a chart at the bottom of the page with values from multiple time series displayed.
  - Start and End date can be specified to load data from different time range
    - Defaults to last week, like observation pages
      - Maybe should default to the last month, to make it clear it that other ranges can be selected?
    - Can be deep linked to from observation pages, so a user could move from viewing an observation to exploring a further time range, and comparing to other platforms
  - For each time series in the chart, include links to download CSV and access ERDDAP (with variables and constraints pre-configured).

Other possible names:
- Explore

Possible enhancements:
- Map layers
  - Radar
  - Forecasts
- Forecast display on the chart for selected platforms

abkfenris:
Monthy/Yearly/... Min/Max/Average table in place/with the chart?

PR #1217, opened 2021-04-16
Explore adding data access and adjustable start date to observations

abkfenris:
Adds a tooltip that can be used to access CSV download, HTML Table, or the ERDDAP dataset for each timeseries on observations pages.

https://user-images.githubusercontent.com/1296209/115042257-4d601d80-9ea1-11eb-93f7-00d655e158f6.mov

Also explored adding a __Start Date__ selector to the same pages, so users can select what range of time they want to display.

image-1

A better way to allow users to explore other time ranges may be to link them to the Compare page if we want to add that #1216 The link could also go in the tooltip.


abkfenris:
A little tweaking of the styles to make the data access links show up better:

image

Issue #1114, opened 2021-03-08
Set color for lines in charts


youngmorse:
The colors are arbitrarily changing on some of the plots. It's most noticeable on the forecast/observation plots and will sometimes change the series color from green to orange while it's reloading. Investigate how we can set the colors so they are fixed by data type and consistent between plots.

abkfenris:
This will take some deeper digging into how Highcharts assigns colors. 

We may want to look at creating a library/cycle of colors that we feed to the charts and be more explicit about what ends up what color. Which is probably best implemented with #1113

abkfenris:
Ok, I've got it so we can pass specific colors to be displayed to the plot, and it will override the default cycle. 

Right now I've got it using NERACOOS blue for the observation, and [a ColorBrewer scheme](https://colorbrewer2.org/#type=qualitative&scheme=Dark2&n=3) for the forecasts.

<img width="1227" alt="image" src="https://user-images.githubusercontent.com/1296209/115064442-ebacad00-9eba-11eb-8439-77246628fee7.png">

Let me know if there is a better set of colors to be using.

abkfenris:
Coloring the tooltips to match the line colors:

<img width="754" alt="image" src="https://user-images.githubusercontent.com/1296209/115067801-67105d80-9ebf-11eb-9245-0656ba32ea49.png">

abkfenris:
Using colors from the NERACOOS visual guide (and fixed styling elsewhere to match).

<img width="1247" alt="image" src="https://user-images.githubusercontent.com/1296209/115080741-42bd7c80-9ed1-11eb-8042-7e09ca3ab038.png">

youngmorse:
This is looking great @abkfenris! It would be good to try and incorporate the NERACOOS color scheme if we can. I'm attaching the neracoos palatte Julianna created as a PDF and I used [coolors.co](https://coolors.co/083d52-178596-2ab78f-ffcd22-ffae6d) to put the palatte online. Maybe use the 'NERACOOS blue' for the obs (which you may be using already), the 'Coastal Meadow' or 'Metallic Seaweed' for the green and see if the rich yellow works in place of the orange you are using? I think the lighter orange in the palatte might be too light, but if Buoy Yellow doesn't work, I think the orange you've got would be ok.
[neracoos palette.pdf](https://github.com/gulfofmaine/Neracoos-1-Buoy-App/files/6327814/neracoos.palette.pdf)

PR #1218, opened 2021-04-18
Update styles including charts and map

### LINKED ISSUE #1113: Change chart line style for observations on forecast plots
On the forecast plots where observations are also available, investigate changing the line type to a dashed line (short or long) to differentiate from the forecast. Related to issue about setting specific colors for plot lines.

abkfenris:
Closes #1113 #1114


abkfenris:
- [ ] Add margin to bottom of platform table.

youngmorse:
From Tom: 
Please implement the Color/Style update for plots. Those look great! Can the update implementation be scheduled to automatically occur at midnight or does this need to be done within work hours?

youngmorse:
Attempt at capturing Tom's comments for the historical record on the PR. I realize I should have included my reply where I told him you usually do deployments after hours but not automagically because they need to be monitored. Thanks for getting the changes up, looks good!

tshyka:
Thanks for getting the changes up! That's good to know that updates are
typically implemented after hours and that they cannot be done
automagically. I certainly do not want to interrupt Alex's or anybody's
sleeping time.

abkfenris:
While I can do updates after hours, there's no schedule. As soon as I tag a commit to release, the pipeline starts, and within about 10 min it goes live. 

Updates are usually an instantaneous cut over, as I believe I've got things set up so the old version doesn't get shut down until the new one is working.

PR #1415, opened 2021-08-10
Change initial bounding box to include Long Island Sound

abkfenris:
Adds an initial bounding box that will display Long Island Sound, instead of using the Gulf of Maine bounding box as the starting view.

New view:

image

Old view:

image

Closes gulfofmaine/NERACOOS-operations#41

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ Additions:
 
 Changes:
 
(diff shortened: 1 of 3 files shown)


tshyka:
The new view looks great. Thanks

abkfenris:
Live now as [0.8.0](https://github.com/gulfofmaine/Neracoos-1-Buoy-App/releases/tag/v0.8.0)

PR #1441, opened 2021-08-20
Catch errors with Sentry.ErrorBoundaries

abkfenris:
We noticed that the CML data is very messy including negative directional degrees. Instead of disabling it right away (I'm about to disable it until we figure out whats up), I first did some work to make sure the Mariner's dashboard can handle when there are errors in data sources better.

Adds a 500 page, and for current conditions cards, or table rows the errors will be caught.

When errors are caught, that section of the page shows that something broke, but tries to still render around the broken element.

image (2)

If errors aren't caught elsewhere in Mariner's Dashboard, then users will be presented a [Sentry feedback form](https://docs.sentry.io/platforms/javascript/guides/react/enriching-events/user-feedback/) and a 500 page explaining that there has been an error.

image (3)

Either way I should still get a notification that something went wrong, and hopefully enough information that I can fix it.


youngmorse (approved):
I think this looks great Alex! It will be more helpful to the users when things go haywire.

PR #2174, opened 2023-01-18
Redirect buoys with NDBC ID in path and show NDBC ID

### LINKED ISSUE #2171: Display NOAA ID when avaliable
So that we can pull the IDs out of the URL (get rid of the spaces), but we'll also need to redirect existing routes.

abkfenris:
If a platform has a NDBC ID it will be shown before the mooring site description.

image

Some tests are likely to fail (and need to get re-setup) until the platforms are adjusted in Buoy Barn to remove the NDBC ID from the platform ID. 

I tested the redirects against a local version of Buoy Barn (hence missing data).

Closes #2171

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ Additions:
 
 Changes:
 
+- Redirect requests to platforms with NDBC ID in path. Closes [#2171](https://github.com/gulfofmaine/Neracoos-1-Buoy-App/issues/2171)
+- Show NDBC ID in platform info. Closes [#2171](https://github.com/gulfofmaine/Neracoos-1-Buoy-App/issues/2171)
 - Use Redux Toolkit.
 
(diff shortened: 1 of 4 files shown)


youngmorse:
Cool! I think this works just fine for to solve the issue and get the URL redirects working. I am wondering if it should be it's own line with **NDBC ID:**  as a header (the way you have Lat/Lon) and then print the actual ID number. But we can table for now, and revisit if we want to be able to add/display other info from Buoy Barn about the mooring at some point.

youngmorse (approved):

abkfenris:
It's quick enough to put it on it's own line, I can do that now.

abkfenris:
With an ID defined:
image

And without:

image

youngmorse:
Looks good! Thanks for making the change @abkfenris!

PR #2645, opened 2023-11-08
transfer models page to next

Adelle-Pitsas:
- Transfers models page to next
   - Adds use-client to layout and wraps layout in `STACProvider`
- Moved the page into a new route group called (modelsMap) since it has a map but isn't the platform map


app/(modelsMap)/models/page.tsx
```diff
@@ -0,0 +1,10 @@
+import { DehydratedPlatforms } from "Features/ERDDAP/hooks/DehydrateComponent"
+import { ModelingPage } from "./modeling"
+
+export default async function ModelingIndexPage() {
+  return (
+    <DehydratedPlatforms>
```
abkfenris, line 6: Do we need `<DehydratedPlatforms>` here? We may want to dehydrate STAC data to speed up catalog loading at some point, but I don't think we have any platform data in this route.

I think `modeling.tsx` could be renamed to `page.tsx` with the exports tweaked.

src/components/NavBar/index.tsx
```diff
@@ -60,13 +60,9 @@ const NeracoosNavBar = () => {
...
-
-            {!isMariners ? (
-              <NavItem>
-                <NavLink href={paths.models}>Model Viewer</NavLink>
-              </NavItem>
-            ) : null}
-
+            <NavItem>
+              <NavLink href={paths.models}>Model Viewer</NavLink>
+            </NavItem>
```
abkfenris, line 65: Right now the model viewer (and maybe some of the servers behind it) aren't quite ready for full public consumption, so we still want to check if the site is at mariners.neracoos.org and hide it from the menu if that is the case. 

Probably the better route (so we avoid a flash as things get rehydrated and update), would be to not show it by default, and only show it if `!window.location.href.includes("://mariners.neracoos.org")`.

abkfenris (changes requested):
Pretty close. I think we have an unneeded layer of hydration happening, and we need to hide the menu item on the public facing site.

app/(modelsMap)/models/page.tsx
Adelle-Pitsas: Great--Took out `<DehydratedPlatforms>` and moved the main `<ModelingPage/>` components into `Page.tsx` to keep auxiliary components out of the main route.

abkfenris (approved):
Feel free to merge when you're happy

Issue #2651, opened 2023-11-09
Modeling Page - Entire page deopted into client-side rendering


Adelle-Pitsas:
Modeling page continuously rerendering on dev. 

https://nextjs.org/docs/messages/deopted-into-client-rendering

"During static rendering, the entire page was deopted into client-side rendering by useSearchParams as there was no [Suspense boundary](https://nextjs.org/docs/app/api-reference/functions/use-search-params#static-rendering) that caught it.

If a route is statically rendered, calling useSearchParams() will cause the tree up to the closest [Suspense boundary](https://nextjs.org/docs/app/api-reference/functions/use-search-params#static-rendering) to be client-side rendered.

This allows a part of the page to be statically rendered while the dynamic part that uses searchParams can be client-side rendered."

Adelle-Pitsas:
Two things are happening here:

1. We are using useSearchParams() which triggered our models page to client-side. The solution was to add a suspense boundary to force it back into SSR. 
2. Since we are forcing the StacMap back to SSR, we are also referencing `document` farther down the tree which isn't accessible from the server, so `useEffect` was used to wait for the initial rendering and then loading the map on the client-side so `document` could be referenced.

PR #2652, opened 2023-11-09
Adelle/fix reload loop

Adelle-Pitsas:
See notes in issue:
Xref issue: https://github.com/gulfofmaine/Neracoos-1-Buoy-App/issues/2651

Fixes rerendering loop by adding a suspense boundary and forcing back to client-side

### DIFF
--- a/app/(modelsMap)/models/modeling.tsx
+++ b/app/(modelsMap)/models/modeling.tsx
@@ -14,7 +14,6 @@ import { useLatestItemsByCollectionIdsQuery, useRootCatalogQuery } from "./stac-
 import { EdrTable } from "./table"
 import { Layer, LoadedData } from "./types"
 
-
 export const ModelingPage = () => {
   const [loading, setIsLoading] = useState(true)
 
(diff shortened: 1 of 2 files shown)


app/(modelsMap)/models/modeling.tsx
```diff
@@ -12,6 +14,31 @@ import { useLatestItemsByCollectionIdsQuery, useRootCatalogQuery } from "./stac-
...
+  }, [])
+
+  return (
+    <>
+      <Row>
+        <Col md={3} style={{ padding: 0 }}>
+          <StacCatalogRoot />
+        </Col>
+        <Col md={9} style={{ paddingLeft: 0 }}>
+          {!loading && <StacMap />}
```
abkfenris, line 31: Is this pattern to delay drawing the map until it's on the client?

abkfenris (approved):
This seems to work when I tested it locally. I'm wondering about the `useState` `useEffect` pattern you have, and what that's supposed to be doing?

app/(modelsMap)/models/modeling.tsx
Adelle-Pitsas: Yes, thats correct

abkfenris: Ah, missed your comment on the issue. Feel free to merge when you are ready.

Issue #2670, opened 2023-11-15
Missing Superlatives 


abkfenris:
Sometimes the [superlatives](https://github.com/gulfofmaine/Neracoos-1-Buoy-App/blob/main/src/Features/ERDDAP/Superlatives/index.tsx#L27) don't show on the home....

<img width="1213" alt="image" src="https://github.com/gulfofmaine/Neracoos-1-Buoy-App/assets/1296209/3784af4c-a8bd-4856-9d4f-7c0c9952dd28">

[platform-no-superlatives.json](https://github.com/gulfofmaine/Neracoos-1-Buoy-App/files/13368508/platform-no-superlatives.json)

abkfenris:
Starting to poke through things in the debugger, `findHighestCondition` keeps finding data, but it is before the 'laterThan' time, so it gets ignored.

<img width="1912" alt="image" src="https://github.com/gulfofmaine/Neracoos-1-Buoy-App/assets/1296209/ba92afef-4a24-42e3-964f-0997b3cd0bb5">

Maybe we need to backoff a few hours if we don't find usable values in the first pass?

abkfenris:
Where the times get compared https://github.com/gulfofmaine/Neracoos-1-Buoy-App/blob/4e0366cb25cc307b158b65770220ec3f5b632e42/src/Features/ERDDAP/Superlatives/index.tsx#L116

In this chunk https://github.com/gulfofmaine/Neracoos-1-Buoy-App/blob/4e0366cb25cc307b158b65770220ec3f5b632e42/src/Features/ERDDAP/Superlatives/index.tsx#L54-L62 I would do something like this to backoff (in Python-ish since that's how I can express more clearly than Typescript).

```python
MAX_BACKOFF_HOURS = 6

for backoff_hour in range(MAX_BACKOFF_HOURS):
    compareTime = laterThan - timedelta(hours=backoff_hour)
    windPlatform, windTimeSeries = findHighestCondition(platforms, compareTime, windSpeed)
 
    if windPlatform:
        break
```

abkfenris:
So it found the wind but not waves. 

<img width="1616" alt="image" src="https://github.com/gulfofmaine/Neracoos-1-Buoy-App/assets/1296209/1eb7b74e-fd54-4560-a139-b7c90e6d34d2">

PR #2672, opened 2023-11-16
Adelle/inconsistent superlatives render

Adelle-Pitsas:
**Functional adjustments**
This PR adjusts how superlatives are looked for:
- Highest wind and biggest waves are looked for conditionally and incrementally up to 6 hours in the past. For example, if a reading is not found in the past hour, it will check in the past two hours. If none found in the past two hours it will search in the past 3, 4, etc hours until it can proclaim a superlative. 
- Superlative info has been put into `useState` and searching for superlatives has been put into `useEffect` for better state management.

**Refactoring**
This PR also refactors the superlatives components and its adjacent functions by separating out how each superlative is rendered. It seeks to avoid the situation where one superlative is not found and neither superlative is rendered.
- Each superlative is now conditionally rendered based on whether or not that superlative exists. This solves the problem of one superlative not being found and neither superlatives being rendered.
- Also included a default in case a superlative cannot be found. It will show  `-- --` (this can be better styled in the future if we want to put time to it). See happy and sad path pictures below.

Closes issue #2670

Happy path -- Superlatives found and displayed correctly:
Screenshot 2023-11-16 at 11 42 39 AM

Sad path -- Superlatives not found:
Screenshot 2023-11-16 at 11 43 16 AM

### DIFF
--- a/src/Features/ERDDAP/Superlatives/index.tsx
+++ b/src/Features/ERDDAP/Superlatives/index.tsx
@@ -12,7 +12,7 @@ import { converter } from "Features/Units/Converter"
 import { UnitSystem } from "Features/Units/types"
 import { paths } from "Shared/constants"
 import { round } from "Shared/math"
-import { calcAnyHourAgoRounded } from "Shared/time"
+import { anHourAgoRounded, hoursBefore } from "Shared/time"
 import { urlPartReplacer } from "Shared/urlParams"
 
 import { usePlatforms } from "../hooks/buoyBarn"
@@ -28,68 +28,13 @@ const windSpeed = new Set(conditions.windSpeed)
 export const Superlatives: React.FunctionComponent = () => {
   const unitSystem = useUnitSystem()
   const { isLoading, data } = usePlatforms()
-  const [platforms, setPlatforms] = useState(data?.features)
 
-  const backOffHours = 6
-
-  //Defaults to the last hour, but will loop through anyway to find the most recent superlative
-  const [windLaterThan, setWindLaterThan] = useState<Date>()
-  const [waveLaterThan, setWaveLaterThan] = useState<Date>()
-
-  useEffect(() => {
-    setPlatforms(data?.features)
-  }, [data])
-
-  //Find time wind superlatives exists
-  useEffect(() => {
-    if (platforms) {
-      for (let hours = 0; hours < backOffHours; hours++) {
-        const backOffTime = calcAnyHourAgoRounded(hours)
-        const { platform: windPlatform, timeSeries: windTimeSeries } = findHighestCondition(
-          platforms,
-          backOffTime,
-          windSpeed,
-        )
-        if (windPlatform) {
-          setWindLaterThan(backOffTime)
-          return
-        }
-      }
-    }
-  }, [platforms])
-
-  //Find time wave superlative exists
-  useEffect(() => {
-    if (platforms) {
-      for (let hours = 0; hours < backOffHours; hours++) {
-        const backOffTime = calcAnyHourAgoRounded(hours)
-        const { platform: wavePlatform, timeSeries: waveTimeSeries } = findHighestCondition(
-          platforms,
-          backOffTime,
-          waveHeight,
-        )
-        if (wavePlatform || waveTimeSeries) {
-          setWaveLaterThan(backOffTime)
-          return
-        }
-      }
-    }
-  }, [platforms])
+  const startTime = anHourAgoRounded()
 
   return (
     <>
-      {data && windLaterThan && waveLaterThan ? (
-        <ShowSuperlatives
-          platforms={data.features}
-          unitSystem={unitSystem}
-          laterThan={
-            windLaterThan && waveLaterThan
-              ? windLaterThan <= waveLaterThan
-                ? windLaterThan
-                : waveLaterThan
-              : windLaterThan || waveLaterThan
-          }
-        />
+      {data ? (
+        <ShowSuperlatives platforms={data?.features} unitSystem={unitSystem} searchStartTime={startTime} />
       ) : (
         <Alert color="primary">Trouble loading platform data</Alert>
       )}
@@ -110,7 +55,7 @@ export const Superlatives: React.FunctionComponent = () => {
 interface ShowSuperlativesProps {
   platforms: PlatformFeature[]
   unitSystem: UnitSystem
-  laterThan: Date
+  searchStartTime: Date
 }
 
 interface HighestCondition {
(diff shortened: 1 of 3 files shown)


Adelle-Pitsas:
Adjusting tests now for the code that I changed

src/Features/ERDDAP/Superlatives/index.tsx
```diff
@@ -144,21 +177,13 @@ interface HighestConditionsProps {
...
+const HighestConditions: React.FunctionComponent<HighestConditionsProps> = ({ platform, timeSeries, unitSystem }) => {
   const dataConverter = converter(timeSeries.data_type.standard_name)
 
   const url = urlPartReplacer(paths.platforms.platform, ":id", platform.id as string)
 
   return (
     <React.Fragment>
-      <Link href={url}>
-        <h6>{title}</h6>
-      </Link>
```
abkfenris, line 161: Initially I was worried about removing this link, then I realized we had another link to the platform already with the platform name.

src/Features/ERDDAP/Superlatives/index.tsx
```diff
@@ -29,17 +30,17 @@ export const Superlatives: React.FunctionComponent = () => {
 
   const lastHour = anHourAgoRounded()
```
abkfenris, line 31: If we're not passing `lastHour` into `<ShowSuperlatives>`, then this can probably get removed.

With a backoff, we probably still want to pass `lastHour` in, and then tweak it to find the backoff time, as that should make it easier to have more predictable tests.

src/Features/ERDDAP/Superlatives/index.tsx
```diff
@@ -51,48 +52,78 @@ interface ShowSuperlativesProps {
...
-      <Card style={{ marginTop: "1rem", marginBottom: "1rem" }}>
-        <CardHeader>
-          <h5>Latest Conditions</h5>
-        </CardHeader>
-
-        <CardBody>
-          <Row>
-            <Col>
+export const ShowSuperlatives: React.FunctionComponent<ShowSuperlativesProps> = ({ platforms, unitSystem }) => {
+  const backoffHours = 6
```
abkfenris, line 56: It might be clearer we pull `backoffHours` out of the function and make it a top of file `BACKOFF_HOURS` constant?

abkfenris (commented):
My quick run through of the code so far looks pretty good. 

I'm adding @youngmorse as a reviewer, as she probably has a better idea what NERACOOS might want for styling when a superlative isn't found.

src/Features/ERDDAP/Superlatives/index.tsx
Adelle-Pitsas: Yeah, seemed extraneous so I thought it would be okay to move it and remove the link

src/Features/ERDDAP/Superlatives/index.tsx
Adelle-Pitsas: Would the tests rely specifically on that `anHourAgoRounded()` function? I created the one that calculates the backoff depending on the number we pass it--would that be not super dependable testing wise?

abkfenris: It's less about the function itself, it's more about where it's used. 

By having `lastHour` (or really the `Date.now()` within it) outside of the component that finds and renders the superlatives, we can set a different `lastHour` when testing, to make sure we match the times of the test data. By having `Date.now()` now being called inside the display component, we now would need to mock it to get the tests to be repeatable.

Adelle-Pitsas: Got it! Thanks for clarifying.

Adelle-Pitsas:
Pushing up fixes for map height-- see comments on issue #2676. 

Closes #2676

app/(platformMap)/layout.tsx
```diff
@@ -16,8 +16,13 @@ export default function Layout({
   bottom: React.ReactNode
   belowMap: React.ReactNode
 }) {
-  const [ref, { height }] = useMeasure<HTMLDivElement>()
   const path = usePathname()
+  const params: { regionId?: string } = useParams()
+  let [ref, { height }] = useMeasure<HTMLDivElement>()
+
+  if (height < 420) {
+    height = 420
```
abkfenris, line 24: Where did you get this value for a minimum height?

src/Features/ERDDAP/Superlatives/index.tsx
```diff
@@ -27,11 +28,12 @@ const windSpeed = new Set(conditions.windSpeed)
...
 
-  const lastHour = anHourAgoRounded()
+  const backOff = 6
+  const backOffTime = calcAnyHourAgoRounded(backOff)
 
   return (
     <UsePlatforms>
-      {({ platforms }) => <ShowSuperlatives laterThan={lastHour} {...{ platforms, unitSystem }} />}
+      {({ platforms }) => <ShowSuperlatives laterThan={backOffTime} {...{ platforms, unitSystem }} />}
     </UsePlatforms>
```
abkfenris, line 37: I think this is duplicating the time in the loops below (74 and 90).

src/Features/ERDDAP/Superlatives/index.tsx
```diff
@@ -56,43 +63,81 @@ export const ShowSuperlatives: React.FunctionComponent<ShowSuperlativesProps> =
...
+        platforms,
+        windBackOffTime.current,
+        windSpeed,
+      )
+      if (windPlatform || windTimeSeries) {
+        setWindSuperlative({ platform: windPlatform, timeSeries: windTimeSeries })
+        return
+      }
+    }
+  }, [platforms, laterThan])
```
abkfenris, line 86: If I'm reading this right, this isn't starting with the `Date` passed in as `laterThan` but still generating a new `Date.now()` in the loop. 

It might be easiest to see this by running the Storybook tests with `yarn storybook` and then http://localhost:6006/?path=/docs/erddap-superlatives--docs will render https://github.com/gulfofmaine/Neracoos-1-Buoy-App/blob/main/src/Features/ERDDAP/Superlatives/index.stories.tsx

In the test I specify two different `laterThan`s to show the difference in how the component will render if it can find readings or not. 

https://github.com/gulfofmaine/Neracoos-1-Buoy-App/blob/76bec6d6faa6e2cba4d6ed9c906379c1760f44f1/src/Features/ERDDAP/Superlatives/index.stories.tsx#L19-L35

The first test should be able to show both of the conditions, and the second should miss both (it would be great to find a time/tweak things to demonstrate what happens if only one is found).

It's pulling the platform test data in from https://github.com/gulfofmaine/Neracoos-1-Buoy-App/blob/main/src/stories/platforms.ts

src/Features/ERDDAP/Superlatives/index.tsx
Adelle-Pitsas: Good catch! Thanks. I'll refactor so theres no duplicate on like 66

src/Features/ERDDAP/Superlatives/index.tsx
Adelle-Pitsas: Thanks for writing all that out. I misunderstood how the test was running.

Just so I don't make another incorrect assumption, would you mind confirming that I am thinking about the issue correctly?

- Because `<ShowSuperlative/>` wasn't finding a superlative in the `laterThan` timeframe, we are wanting to incrementally back up that time until we find a reading that can be a superlative. 
- We don't want this loop logic in the `<ShowSuperlatives />` component for testing's sake because we want to pass in a stubbed time to `<ShowSuperlatives/>` that reflects the stubbed data that will give us the above happy (superlatives found) and sad (no superlatives found) paths.

app/(platformMap)/layout.tsx
Adelle-Pitsas: Anything less than 420 cuts off buoys on the map with the current Maine-centered zoom.

src/Features/ERDDAP/Superlatives/index.tsx
abkfenris: I think we can have the backoff logic in `<ShowSuperlatives/>`, it just shouldn't have the time (the `Date.now()` call) that we are starting our search with in it, so that there is at least some separation between state and display. 

State in this case is the current selection of `platforms` and `laterThan` (`Date.now()`) that starts the search. In theory by splitting the state and the display, we can make sure that if we always pass the same inputs into `<ShowSuperlatives/>` then we will get the same outputs.

Adelle-Pitsas: Yes, definitely agree about the separation of state and display. 

I'm not following how its possible to do the superlatives logic in this component with the testing constraints you've laid out. To get the superlatives we have to pass a date as an arg into the  `findHighestCondition()` function and if we are looping through to back off the date, that date will have to change to determine whether superlatives exist in that timeframe.

I have a local branch going that refactors it so its parent component does the logic to 1) find the most recent time there are superlatives, and 2) passes just that date into the `<ShowSuperlatives/>` . This way the already predetermined date is being passed into the component and then doesn't change once its in the `<ShowSuperlatives/>` component.

Adelle-Pitsas:
Okay, changes have been made based on Alex's and my conversation yesterday/late last week.

Problem: State and display should be kept in separate components where parent component handles state and child handles display. State that is passed down to superlatives should be: `platforms`, `unitSystem`, and `laterThan`. `laterThan` should not be changed with `Date.now()` in the display component to avoid having to stub more data for tests.

Current implemented solution: Parent component is refactored so it is no longer using deprecated `RenderProps` and just calls the custom hook to get platforms. Once platforms have loaded, it will look through for the most recent hour that both superlatives exist. It will then pass that hour (as `laterThan`). I considered passing a starting `laterThan` time and adjusting the time in the ShowSuperlatives component but that ended up being pretty messily done. This seems like a good separation of state and display and more readable. This solution passes all current written tests.

Adelle-Pitsas:
Also zoom was broken in this branch, so I fixed it back to what it currently is in dev. It has some funkiness to it still but I'm going to create a different issue and branch for that

PR #2685, opened 2023-11-22
Adelle/map zoom bugs

### LINKED ISSUE #2683: Zoom out jumps back to initial view
I might have unknowingly broken this in one of my first branches so it exists on `dev`. Whenever view is set, zooming out will jump back to set view. I believe this has to do with how we are setting the home and region view manually.

https://github.com/gulfofmaine/Neracoos-1-Buoy-App/assets/108096652/997a6aff-b479-4f4f-ab1c-77855e2495dc

Adelle-Pitsas:
This PR seeks to fix a few map bugs that (I believe) were caused by the map state and display not being in sync. 
Issues:
- Inability to zoom manually on map because view was hardcoded (see issue #2683, closes #2683 )
- Region zoom wouldn't work on a hard refresh (ex: a hard refresh on `/region/GREAT` would set map back to initial view)
- Sometime the map wouldn't recognize when the region was changed (this is the corner case where changing regions rapidly would lead to the map view not changing). I've tested this manually several times and I haven't gotten it to break but please test this and let me know if its still breaking for on your end @abkfenris .

The main changes in this branch is taking out the redux handling of the map view and putting it in local state in the `ErddapMapBase` component. From there, when params change, boundingBox is set which triggers a set of the `RMap` view state, which triggers a change in the component's local state, which triggers a rerender of the display in alignment with `RMap`

### DIFF
--- a/src/Features/ERDDAP/Map/index.tsx
+++ b/src/Features/ERDDAP/Map/index.tsx
@@ -9,14 +9,12 @@ import { fromLonLat, transformExtent } from "ol/proj"
 import { Button } from "reactstrap"
 import { RFeature, RLayerVector, RMap, RPopup, RStyle } from "rlayers"
 
-import { useStatefulView } from "Features/StatefulMap"
 import { colors } from "Shared/colors"
 import { paths } from "Shared/constants"
-import { BoundingBox, regionList } from "Shared/regions"
+import { BoundingBox, regionList, regions } from "Shared/regions"
 import { urlPartReplacer } from "Shared/urlParams"
 import { EsriOceanBasemapLayer, EsriOceanReferenceLayer } from "components/Map"
 import { useCallback, useEffect, useLayoutEffect, useMemo, useRef, useState } from "react"
-import type { RView } from "rlayers/RMap"
 
 import { aDayAgoRounded } from "Shared/time"
 import { useParams } from "next/navigation"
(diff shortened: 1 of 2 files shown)


abkfenris:
> The main changes in this branch is taking out the redux handling of the map view and putting it in local state in the `ErddapMapBase` component.

Can we rip out all of `src/Features/StatefulMap` then? And remove it from the redux store? 

Then we would be pretty close to being able to remove redux entirely as only `src/Features/Units` would still depend on it, and I think that could reasonably be replaced by `useContext` and `useState` or possibly cookies.

abkfenris:
Hmm, it looks like it is letting the user change the view, but if you zoom, and then select a platform it jumps back out to the initial view again.

Adelle-Pitsas:
I'll rip out all of `src/Features/StatefulMap` and make sure nothing is broken by it. Not familiar with removing from Redux store but I'll do a wee bit of research. 


> Hmm, it looks like it is letting the user change the view, but if you zoom, and then select a platform it jumps back out to the initial view again.

Thanks for catching this! I'll work on a fix.

abkfenris:
I think the import and redux slice function can be removed from https://github.com/gulfofmaine/Neracoos-1-Buoy-App/blob/main/app/store.ts would do it. I don't think anything else imports from StatefulMap at this point.

Adelle-Pitsas:
Newest commits fix the selected view
- Map maintains regional view when platform is selected from region
- Platform selection on map only persists on specific platform view

Also rips out `statefulMap` completely

https://github.com/gulfofmaine/Neracoos-1-Buoy-App/assets/108096652/c4988527-6a8a-4b32-aa84-5079a2e46836

abkfenris (approved):
I think this looks good from my playing around with it, and I'm glad to see the stateful map go!

We can probably do another PR, but I noticed the map order on mobile for `/` is off.

Issue #2691, opened 2023-11-29
Some routes not working on refresh or direct navigation


Adelle-Pitsas:
I've noticed some platform routes don't work on refresh or on direct navigation -- would be an issue if someone was trying to access a specific url (say [https://mariners-dev.aws.neracoos.org/platform/GREAT_BAY/observations/sea_water_salinity](https://mariners-dev.aws.neracoos.org/platform/GREAT_BAY/observations/sea_water_salinity) and it led them to our error page.)

Some routes I've noticed it on: 
- Any observation routes [https://mariners-dev.aws.neracoos.org/platform/BHBM3/observations/air_pressure_at_sea_level](http://localhost:3000/platform/BHBM3/observations/air_pressure_at_sea_level)
- Any forecast routes [https://mariners-dev.aws.neracoos.org/platform/BHBM3/forecast/air_temperature](http://localhost:3000/platform/BHBM3/forecast/air_temperature)

abkfenris:
Good catch. Now I remember encountering that while getting the initial routing working, but I guess I forgot to continue testing for it. 

Probably would be a good set of Playwright tests.

abkfenris:
Do you mean going to say https://mariners-dev.aws.neracoos.org/platform/E01 on a fresh load or reload? While it would be nice to have a more focused zoom on the platform in that case, we're not doing that currently so it shouldn't be a blocker (at least compared to inaccessible routes above). 

Or do you mean keeping the zoom from pre-reload? I wouldn't worry about persisting zoom state across full reloads.

Adelle-Pitsas:
Its the latter--Sounds good. I worry about saving zoom state when fully reloading

Adelle-Pitsas:
Update: So this is a hydration error where a full refresh would attempt to rerender some components on the page, but they didn't match what was rendered on the server thus throwing this error: 
`Uncaught Error: Hydration failed because the initial UI does not match what was rendered on the server.`
Taking out the `DehydratedPlatforms` wrapper in `/app/(platformMap)/@sidebar/platform/[platformId]/page.tsx` removes the error and allows the page to render--but this solution only works with `yarn dev` so still working on a solution that works with a build step.

Adelle-Pitsas:
[...]

PR #2695, opened 2023-12-01
Adelle/refresh navigation bug

Adelle-Pitsas:
This addresses https://github.com/gulfofmaine/Neracoos-1-Buoy-App/issues/2691 where observation and forecast routes were hitting our error page on hard refresh.

Changes made:

Removed DehydratedPlatforms wrapper and added a client check to make sure that components using platform information was the same on client and server.

### DIFF
--- a/app/(platformMap)/layout.tsx
+++ b/app/(platformMap)/layout.tsx
@@ -1,6 +1,7 @@
 "use client"
 import { useParams, usePathname } from "next/navigation"
 import * as React from "react"
+import { useEffect, useState } from "react"
 import { useMeasure } from "react-use"
 import { Col, Row } from "reactstrap"
 import { ErddapMap } from "../../src/Features/ERDDAP/Map"
@@ -21,6 +22,11 @@ export default function Layout({
   const params: { regionId?: string; platformId?: string } = useParams()
   let [ref, { height }] = useMeasure<HTMLDivElement>()
   const platformId = params.platformId
+  const [isClient, setIsClient] = useState(false)
+
+  useEffect(() => {
+    setIsClient(true)
+  }, [])
 
   if (height < 420) {
     height = 420
@@ -48,7 +54,7 @@ export default function Layout({
         </Col>
       </Row>
 
-      {(isPlatformView && bottom) ?? <Row>{bottom}</Row>}
+      {(isPlatformView && isClient && bottom) ?? <Row>{bottom}</Row>}
     </React.Fragment>
   )
 }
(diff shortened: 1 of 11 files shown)


app/(platformMap)/layout.tsx
```diff
@@ -48,7 +54,7 @@ export default function Layout({
         </Col>
       </Row>
 
-      {(isPlatformView && bottom) ?? <Row>{bottom}</Row>}
+      {(isPlatformView && isClient && bottom) ?? <Row>{bottom}</Row>}
```
abkfenris, line 57: Why do we only need to check for client for the bottom content?

src/Features/ERDDAP/Map/index.tsx
```diff
@@ -170,11 +170,16 @@ export const ErddapMapBase: React.FC<BaseProps> = ({ platforms, platformId, heig
...
+  const { isLoading, data } = usePlatforms()
+  const [isClient, setIsClient] = useState(false)
+
+  useEffect(() => {
+    setIsClient(true)
+  }, [])
+
+  if (data?.features && isClient) {
+    return <ErddapMapBase platforms={data?.features} platformId={platformId} height={height} />
+  }
```
abkfenris, line 182: Do we need to `return null` as a drop through case?

abkfenris (commented):
Any chance you could add some Playwright tests that test if the pages can be directly loaded? It definitely seems like the kind of thing we (or at least I) would be likely to unknowingly break again in the future.

src/Features/ERDDAP/Map/index.tsx
Adelle-Pitsas: Yes, thanks for catching this

app/(platformMap)/layout.tsx
abkfenris: Can we push client only rendering down to the chart components themselves?

abkfenris: I've got [`'use client'` on at least some of the charts](https://github.com/gulfofmaine/Neracoos-1-Buoy-App/blob/main/src/components/Charts/LargeTimeSeries.tsx#L1), so it may be something in between.

It looks like there may be `async` missing on some of the `page.tsx` at this level https://github.com/gulfofmaine/Neracoos-1-Buoy-App/blob/30dbb04393f6359314ec057c77fc8e409f6c5f69/app/(platformMap)/%40bottom/platform/%5BplatformId%5D/observations/%5BstandardName%5D/page.tsx#L5

abkfenris:
I swear, I'm trying to get there.

Adelle-Pitsas:
No rush at all!!

abkfenris (approved):
Looks good to me! Thank you!

PR #2699, opened 2023-12-05
Adelle/models page cleanup

Adelle-Pitsas:
Just threw in a few design edits to the models page:
- Adjusted placement of loading signal so it was in line with what it was loading
- Adjusted loading messages to signal what they were loading
- Compartmentalized the variable selection on the left scrolling sidebar so it was intuitive to scroll
- Adjust spacing generally

Models landing page before:
Screenshot 2023-12-05 at 4 11 07 PM

Models landing page after:
Screenshot 2023-12-05 at 4 15 19 PM

Data loading (bottom) before:
Screenshot 2023-12-05 at 4 15 51 PM

Data loading (bottom) after:
Screenshot 2023-12-05 at 4 16 25 PM

### DIFF
--- a/app/(modelsMap)/models/modeling.tsx
+++ b/app/(modelsMap)/models/modeling.tsx
@@ -111,12 +111,24 @@ export const TableChart = () => {
   } else if (catalogQuery.isLoading) {
     return <div style={{ textAlign: "center" }}>Loading catalog...</div>
   } else if (point) {
-    return <Alert color="warning">Layer needs to be selected to display info for point</Alert>
+    return (
+      <Alert color="warning" style={{ marginLeft: "10px" }}>
+        Layer needs to be selected to display info for point
+      </Alert>
+    )
   } else if (currentLayer.id) {
-    return <Alert color="warning">Point needs to be selected to display info for current layer</Alert>
+    return (
+      <Alert color="warning" style={{ marginLeft: "10px" }}>
+        Point needs to be selected to display info for current layer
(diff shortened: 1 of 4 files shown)


app/(modelsMap)/models/modeling.tsx
```diff
@@ -24,15 +24,23 @@ export const ModelingPage = () => {
...
-      <Row>
-        <Col md={3} style={{ padding: 0 }}>
+      <Row style={{ paddingLeft: "10px" }}>
+        <Col
+          md={3}
+          height="55vh"
+          style={{
+            padding: "5px",
+            border: "1px solid #0000002d",
+            borderRadius: "10px",
```
abkfenris, line 34: It may just be me, but for some reason it doesn't look right when catalog listings next to maps are rounded, in the way that is ok for info boxes. [Generally](http://nextgen-dev.ioos.us/xreds/) catalogs have a bit less styling because they are info dense

abkfenris (approved):
Generally I think it looks good, but a few things to think about.

app/(modelsMap)/models/modeling.tsx
Adelle-Pitsas: Sounds good! Yeah, don't 100% know the usual design standards, just mimicking what I've seen on the rest of mariners and wanting to create some spacing so its easier to digest when first arriving at the page. I'll remove the roundedness for the catalog!

PR #2716, opened 2023-12-19
patch fix for removing old forecast display

Adelle-Pitsas:
This is a patch fix for forecasts displays ([issue #166](https://github.com/gulfofmaine/NERACOOS-operations/issues/166) in Neracoos Operations). Checks to make sure that forecast data is not far in the past. This was fixed in buoy barn already, but these changes can act as another check in case something like this happens again.

### DIFF
--- a/src/Features/ERDDAP/Platform/Forecasts/Page/index.tsx
+++ b/src/Features/ERDDAP/Platform/Forecasts/Page/index.tsx
@@ -2,24 +2,24 @@
 /**
  * Load and display forecasts
  */
-import { FontAwesomeIcon } from "@fortawesome/react-fontawesome"
 import { faInfoCircle } from "@fortawesome/free-solid-svg-icons"
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome"
 import { Point } from "@turf/helpers"
 import React from "react"
-import { Alert, Row, Col, Tooltip } from "reactstrap"
+import { Alert, Col, Row, Tooltip } from "reactstrap"
 
-import { MultipleLargeTimeSeriesChartCurrent } from "components/Charts/MultipleLargeTimeSeriesCurrent"
-import { colorCycle } from "Shared/colors"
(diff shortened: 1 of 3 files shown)


Adelle-Pitsas:
Playwright tests failing because units test rely's on platform E01 which doesn't have any recent data

src/Features/ERDDAP/Platform/Forecasts/Page/index.tsx
```diff
@@ -88,18 +87,24 @@ export const Forecast = ({ platform, forecast_type, ...props }: Props) => {
...
-        url: meta.source_url,
-        dashStyle: "Solid",
-        color: colorCycle[index + 1],
-      })
-    }
-  })
+  forecastResults
+    .filter((r) => {
+      if (r.result.data) {
+        return r.result.data[0].time > threeDaysAgoRounded()
```
abkfenris, line 93: I think we might want to check the most recent time, rather than the first time in the forecast. 

FVCOM is kinda screwy in that it's a five day 'forecast' but it's started 2 days back. So the earliest day can still 4 days ago, but there is still a day that's ahead.

Adelle-Pitsas: Just coming back to this given the issue in buoy barn this morning. 

I'm checking the first time/index in the forecast/array so I don't include the forecasts from Oct (iterating through and not using any that are more than 3 days ago). Given that info about FVCOM having the earliest start day be max 4 days ago, it seems like I should still check the first time in the forecast but instead of `threeDaysAgo` it should be at least four days ago, so include any forecast data that is within the last four days.

I'm not 100% sure what we'd be checking if we were to use the most recent time in the forecast/last index in the array as those are the dates are the same in both forecasts models.

Adelle-Pitsas:
As of 4/9/24 there seems to be backend support for this. Going to close.

PR #2789, opened 2024-03-22
Adelle/predicted tides

Adelle-Pitsas:
Adds predicted tides to the graph

https://github.com/gulfofmaine/Neracoos-1-Buoy-App/assets/108096652/33002cdb-06bf-402c-9b71-dbaddaeaf814

Still trying to figure out how to extend graph so it shows future predictions
Time picker functionality is _clunky_ --working on a better solution for this


src/Features/ERDDAP/hooks/tabledap.ts
```diff
@@ -16,7 +16,7 @@ import { defaultQueryConfig } from "./hookConfig"
 
 const FORBIDDEN = "Forbidden"
 
-export const getDataset = (timeSeries: PlatformTimeSeries, startTime?: Date) => {
+export const getDataset = (timeSeries: PlatformTimeSeries, startTime?: Date, endTime?: Date) => {
   return async () => {
     const url = urlBuilder([timeSeries], startTime)
```
abkfenris, line 21: The `endTime` likely needs to make it's way into `urlBuilder`, and for `urlBuilder` to also know what it can do with that.

Adelle-Pitsas: Yes, thank you! I was mid way through trying to add end time in when I initially pushed that up. 

Because Tom is wanting to be able to see the changes in the dev server, I think its not in the scope of this PR (i.e. I don't think I will have time) to add the predicted tides in the future. However, I did add some changes so the skeleton of being able to see future predicted tides is built out--just not functional yet. I've updated the current screenshot.

youngmorse (approved):
Looks great Adelle! It's good to have the projected tides for comparison to observations even if we don't have future tides yet.

PR #2793, opened 2024-03-25
add endTime to predicted tides

Adelle-Pitsas:
This adds projected tides in the future to the graph. Default is 1 week in the future, but user can select 2 weeks out from drop down.

The time selection is really clunky so I think its in the scope of this branch to add a native time/date picker instead of the predetermined times from the drop down. So I'll keep this as a draft PR and update as I make progress.

Current screenshot of page with predicted tides:
Screenshot 2024-03-25 at 4 11 16 PM

Updated screenshot of page with predicted tides and date picker:
Screenshot 2024-03-26 at 3 09 31 PM


src/Features/ERDDAP/hooks/tabledap.ts
```diff
@@ -56,10 +56,12 @@ const erddapService = (process.env.NEXT_PUBLIC_ERDDAP_SERVICE || "https://buoyba
 
 export function urlBuilder(timeSeries: PlatformTimeSeries[], startTime?: Date, endTime?: Date): string {
   startTime = startTime ?? aWeekAgoRounded()
+  endTime = endTime ?? weeksInFuture(1)
 
   const constraints = {
     ...timeSeries[0].constraints,
     "time>=": startTime.toISOString(),
+    "time<=": endTime.toISOString(),
```
abkfenris, line 64: Could we only set `time<=` when `endTime` is specified, or at least make sure it's reasonably rounded? That will allow the upstream servers to cache better.

Adelle-Pitsas: Yes, makes sense. Updated

youngmorse (approved):
looks good!

abkfenris (approved):

youngmorse (approved):

Adelle-Pitsas:
A couple more things to do on this before merging:

- [ ] When the sensor doesn't have predicted tides the url stays the same while the endDate and graph show only to today. Make it so the url tracks the correct end date
- [ ] Track all changes with milliseconds so it can handle a `startDate` and `endDate` as the same day (and just plot one days worth of data)

Adelle-Pitsas:
Updated userflow
Things to notice: url changes based on date selection. If the sensor does not have predicted tides/info, the url will redirect to today's date to get the latest info.
Beginning date is beginning of that day, end date is at the end of the specified day

https://github.com/gulfofmaine/Neracoos-1-Buoy-App/assets/108096652/af174144-54c0-40ee-a5af-4cc985b4217e

app/(waterLevel)/water-level/sensor/[sensorId]/[startTime]/[endTime]/[datum]/page.tsx
Comment (abkfenris) [inline app/(waterLevel)/water-level/sensor/[sensorId]/[startTime]/[endTime]/[datum]/page.tsx:1]: It feels like `startTime`, `endTime`, and `datum` should probably be search params rather than path elements.

Comment (Adelle-Pitsas) [reply app/(waterLevel)/water-level/sensor/[sensorId]/[startTime]/[endTime]/[datum]/page.tsx:1]: Yep, 100% right. I completely forgot about search params 🤦🏻‍♀️

Comment (Adelle-Pitsas) [reply app/(waterLevel)/water-level/sensor/[sensorId]/[startTime]/[endTime]/[datum]/page.tsx:1]: Will edit in another branch

PR #2798, opened 2024-03-29
get working model of color changing map up and running

Adelle-Pitsas:
This adds conditional color to the map points based on whether or not that sensor detects notable water levels.
Screenshot 2024-04-01 at 12 08 26 PM

Screenshot 2024-04-01 at 12 09 09 PM

### DIFF
--- a/src/Features/ERDDAP/utils/waterLevelThresholds.tsx
+++ b/src/Features/ERDDAP/utils/waterLevelThresholds.tsx
@@ -36,11 +36,12 @@ export const getSurpassedThreshold = (value, thresholds) => {
 }
 
 export const floodLevelThresholdsAlertColors = (alert, old, opacity) => {
+  console.log(opacity)
   const colorMap = {
-    Major: colors.alertMajor,
-    Moderate: colors.alertModerate,
-    Minor: colors.alertMinor,
-    Action: colors.alertAction,
+    Major: `${colors.alertMajor}${opacity}`,
+    Moderate: `${colors.alertModerate}${opacity}`,
+    Minor: `${colors.alertMinor}${opacity}`,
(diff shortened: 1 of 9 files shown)


src/Features/ERDDAP/utils/waterLevelThresholds.tsx
```diff
@@ -0,0 +1,51 @@
...
+      return value >= t.min_value
+    }
+    return value >= t.min_value && value <= t.max_value
+  })
+  if (range) {
+    return range.name
+  }
+  return null
+}
+
```
Adelle-Pitsas, line 37: Theoretically, a way this can be calculated differently is by using an iterative recursive function. Iterating through the flood levels, and returning the flood level where the value is above the min_value but below the next thresholds min_value. However, this would assume the flood thresholds are in order within the array. This current way seems verbose, but also protects against out-of-order flood thresholds.

Adelle-Pitsas:
I'm not a fan of the colors I chose, so going to revisit those to make it look less chaotic

youngmorse:
This is looking great! I think we can fine tune the colors and marker shape/size once it feels like it's working how we want it too. I agree with Alex, we want to be consistent with NOAA's colors. But maybe we can take some artistic license and tone down the shade/intensity of electric green, fire alarm red, and blaze orange.

youngmorse (approved):

Adelle-Pitsas:
Great. Thanks for all that feedback! Ended up just copying the colors which we can change later.
Other things:
- Changed `alert` names to `threshold` so we don't go crazy
- Added a makeshift legend in the map
- Changed the plot band colors on the graph accordingly
Screenshot 2024-04-02 at 10 59 07 AM

PR #2800, opened 2024-04-03
Adelle/mobile orientation

### LINKED ISSUE #2808: Make sure its easily viewable on mobile

Adelle-Pitsas:
**The current mobile orientation views**
A few notes:
- On the detail page: If the user is on mobile, the graph will default to a 3 day span (Instead of 3 days in the past and a week in the future, the view is one day in the past and two days in the future). I can easily change this if that's not the correct timespan that is useful
- I don't love the current detail page order. I'm thinking it should have the global sensor info first, the graph and all its selectors second, and the map last but would love thoughts/feedback on this before I sink a chunk of time in html land.

`/water-level` list page on laptop:
Screenshot 2024-04-04 at 10 54 57 AM
`/water-level` list page on mobile:
Screenshot 2024-04-04 at 10 55 52 AM

Water level sensor detail page on laptop:
Screenshot 2024-04-04 at 10 56 36 AM
Water level sensor detail page on mobile:
Screenshot 2024-04-04 at 10 57 30 AM
Screenshot 2024-04-04 at 11 03 34 AM

closes #2808


abkfenris (approved):
>     * I don't love the current detail page order. I'm thinking it should have the global sensor info first, the graph and all its selectors second, and the map last but would love thoughts/feedback on this before I sink a chunk of time in html land.

I agree, that feels more natural. I'd maybe move the explicit platform selector below the map to make it more clear that they are linked, and the datum selector into the sensor info block, similar to the unit selector.

Adelle-Pitsas:
Sounds good. I'll play around with it. Tying @youngmorse in on this too since it diverges from the original wireframe she drew up.

> I'd maybe move the explicit platform selector below the map to make it more clear that they are linked

The only thing I'm thinking of with this is that they are both sensor selection components and it makes more sense to me as a user to have them in two different places on the page for ease of changing sensors regardless of where they are on the page. I'm thinking they don't need to be near each other for the user to understand they have essentially the same functionality.

abkfenris:
> > I'd maybe move the explicit platform selector below the map to make it more clear that they are linked
> 
> The only thing I'm thinking of with this is that they are both sensor selection components and it makes more sense to me as a user to have them in two different places on the page for ease of changing sensors regardless of where they are on the page. I'm thinking they don't need to be near each other for the user to understand they have essentially the same functionality.

I'm more thinking we probably don't need both, or only need to show the selector if the map can't render. At least to me, it also makes things look a bit lopsided where it is now, and there are 4+ different areas of the page where there are options to change the current view, so consolidating those might help things.

app/(waterLevel)/water-level/sensor/[sensorId]/[startTime]/[endTime]/[datum]/page.tsx
```diff
@@ -29,27 +29,46 @@ export default function SensorIdPage({ params }) {
   }, [data])
 
   return (
-    <Row>
-      <Col style={{ width: "20vw" }}>
-        <div style={{ width: "100%" }}>
+    <div>
+      {window.innerWidth < 576 ? (
+        <Col>
```
Comment (Adelle-Pitsas) [inline app/(waterLevel)/water-level/sensor/[sensorId]/[startTime]/[endTime]/[datum]/page.tsx:34]: This is pretty ugly and is a very manual way to set the order of the components for mobile vs non-mobile. Because its not a perfect grid (i.e. the bottoms/tops don't align in side/side view) the sidebar components have to be wrapped in the large view but not in the mobile view since the `platformInfo` and map components need be be separate to add the graph between. 
Feedback absolutely welcome @abkfenris  but not sure how much more its worth fudging around with it.

Comment (abkfenris) [reply app/(waterLevel)/water-level/sensor/[sensorId]/[startTime]/[endTime]/[datum]/page.tsx:34]: I think this could be done with [Bootstrap/Reactstrap breakpoint classes](https://getbootstrap.com/docs/5.3/layout/grid/#css). The other option might be `useMeasure` or one of the other utilities in `react-use`. https://streamich.github.io/react-use/?path=/story/sensors-createbreakpoint--docs

Comment (Adelle-Pitsas) [reply app/(waterLevel)/water-level/sensor/[sensorId]/[startTime]/[endTime]/[datum]/page.tsx]: Thanks for sending these over. I looked at them and they don't seem to solve the specific problem of what we have going on, that is, conditionally rendering a container in order to avoid conditionally rendering the contents twice. I did add `useBreakpoint` to avoid the use of calling `window.innerWidth`.

PR #2811, opened 2024-04-09
Adelle/selector consolidation

Adelle-Pitsas:
This branch is attempting to consolidate url changing components to the sidebar so there is a main place where a user can customize their page view.

Changes:
- Moves the sensor selector to sidebar
- Moves datum selector to dropdown and change component to a radio button group instead of another dropdown

See comments from closed PR #2800 

https://github.com/gulfofmaine/Neracoos-1-Buoy-App/assets/108096652/3f6cae31-4902-490d-92b3-af65fc013db6

### DIFF
--- a/app/(waterLevel)/water-level/sensor/[sensorId]/[startTime]/[endTime]/[datum]/page.tsx
+++ b/app/(waterLevel)/water-level/sensor/[sensorId]/[startTime]/[endTime]/[datum]/page.tsx
@@ -5,13 +5,13 @@ import { ErddapWaterLevelMapBase } from "Features/ERDDAP/waterLevel/map"
 import { WaterLevelObservationContent } from "Features/ERDDAP/waterLevel/observationContent"
 
 import { filterForSensors } from "Features/ERDDAP/waterLevel/sensor"
-import { PlatformInfo } from "components/PlatformInfo/platformInfo"
 
 import { fromLonLat } from "ol/proj"
 import { useEffect, useState } from "react"
 import { Col, Row } from "reactstrap"
 import { useDecodedUrl } from "util/hooks"
 import { createBreakpoint } from "react-use"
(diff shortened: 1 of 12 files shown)


abkfenris (approved):
The layout looks much cleaner! 

Any reason why you went with a radio group rather than a dropdown our a [Button Group](https://6-4-0--reactstrap.netlify.app/components/button-group/)?

Adelle-Pitsas:
The radio input group was discussed as potentially more preferable than the drop down when we had our meeting with Tom and Anna. Also thinking that button groups are a bit nonconventional when we have more than a couple of options. But can easily be changed back depending on what they/we prefer!

abkfenris:
Ok. I was thinking if a button group was the call, then it would probably need to be acronyms with the full name on hover for size.

It might be worth showing them at least a mockup of how it would look with each now that they are stacked up, as right now there are three different methods of inputting  what are really 3 single selections on top of each other.

youngmorse:
I think this is good for the next update! Having the different types of selection options in the working version will help get to a preference like Alex is suggesting without having to do a mockup for each selection type. Though it does make the date selector seem isolated now that it's the only one on the main block. 

I'm also wondering about doing something more like [CO-OPs](https://tidesandcurrents.noaa.gov/waterlevels.html?id=8418150&units=standard&bdate=20240408&edate=20240409&timezone=GMT&datum=MHHW&interval=6&action=) where the selection items are grouped together and called out with the shaded background. Thoughts for a future iteration!

Adelle-Pitsas:
Yeah, I agree the timeframe selector looks a little weird by itself down below, though it is directly adjusting the graph so I think it makes sense where it is. Might be worth wrapping the graph and timeframe in their own box for clarity/cleanness. I also like the idea of all the selectors in their own area like CO-OPs. I'll jot some stuff down and make a note in our water-level page issue.

PR #2850, opened 2024-04-15
Use station name for styling when it exists

abkfenris:
We now have an optional field `station_name` in the API, so when it's available we should include it in how we display the name of stations, and allow the mooring site description to be a stand alone longer text field.

When a platform has a defined station name it shows both the ID and the name:

image

But it can fall back to just the ID

image

Other places with the name displayed:
image

image

image


Adelle-Pitsas (approved):

youngmorse (approved):
Looks good!

Issue #2864, opened 2024-04-22
Fix datum calculation


Adelle-Pitsas:
Datum offset is inverted

Adelle-Pitsas:
Looking into this more fully and it looks like it is an issue with both how the datum is used (should be - instead of +) and it is not accurately being converted from meters.

PR #2870, opened 2024-04-23
fix datum calculation and fix y-axis

Adelle-Pitsas:
This fixes the datum calculation (issue described and updated in issue #2864)

This also fixes the y axis issue where the y-axis was too big for some timeseries (mostly MLLW)

Comparison to Tides&Currents as a test (within rounding error):
Screenshot 2024-04-23 at 4 22 34 PM

Screenshot 2024-04-23 at 4 23 16 PM

### DIFF
--- a/src/Features/ERDDAP/waterLevel/chart/chartDisplay.tsx
+++ b/src/Features/ERDDAP/waterLevel/chart/chartDisplay.tsx
@@ -9,6 +9,7 @@ import { LargeTimeSeriesWaterLevelChart } from "./largeTimeSeriesChart"
 import { getDatumDisplayName } from "Shared/dataTypes"
 import { TimeframeSelector } from "../timeframeSelector"
 import { displayShortIso, getIsoForPicker, manuallySetFullEODIso, roundDate, shortIso } from "Shared/time"
+import { round } from "@turf/helpers"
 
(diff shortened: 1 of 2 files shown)


abkfenris (approved):

youngmorse (approved):

src/Features/ERDDAP/waterLevel/chart/chartDisplay.tsx
```diff
@@ -49,14 +51,21 @@ export const WaterLevelChartDisplay: React.FunctionComponent<ChartTimeSeriesDisp
           acc[level.name] =
             level.name === "Major"
               ? {
-                  minValue: dataConverter.convertToNumber(level.min_value, unitSystem) + datumOffset,
-                  maxValue: dataConverter.convertToNumber(level.min_value, unitSystem) + 1 + datumOffset,
+                  minValue:
+                    dataConverter.convertToNumber(level.min_value, unitSystem) -
```
abkfenris, line 55: It might make sense to swap the order of things here, and maybe capture the datum offsetting in it's own function so that we can test it (map MHHW it over an array, ask for the max and see that it's near 0, map MLLW over an array, get the min, see that it's near 0). Then apply the converter to the output of the already offset value.

PR #2899, opened 2024-05-09
persist unit selection on page refresh and new platform/sensor selection

### LINKED ISSUE #2858: Units should stay the same when changing stations
- [ ] Unit system selection reverts back to English from Metric when you change stations or change time range. Works okay with Datum selection change (hard to show in a screen shot but easy to replicate!)
Screenshot 2024-04-12 at 4 20 53 PM
Screenshot 2024-04-12 at 4 21 02 PM

_Originally posted by @youngmorse in https://github.com/gulfofmaine/Neracoos-1-Buoy-App/issues/2701#issuecomment-2052472850_

Adelle-Pitsas:
This adds "unit" to `sessionStorage` along side `redux` so unit persists on refresh and platform selection when on the same tab. If another tab/window/session is opened, the unit selection will default back to initial state (English units).

Meant to make this a draft PR, whoops. I'm not sure we want units in session storage for the entire app or just for the water level portion. Seems like this universal change won't be super intrusive for how I understand most users use the buoy dashboard, but happy to receive feedback on this.

closes #2858

### DIFF
--- a/app/store.ts
+++ b/app/store.ts
@@ -2,10 +2,39 @@ import { configureStore } from "@reduxjs/toolkit"
 
 import { unitReducer as unit } from "Features/Units"
 
+const preloadedState = loadFromLocalStorage()
+
 export const store = configureStore({
   reducer: {
     unit,
   },
+  preloadedState: {
+    unit: preloadedState,
+  },
+})
+
+store.subscribe(() => {
+  saveToLocalStorage(store.getState().unit)
 })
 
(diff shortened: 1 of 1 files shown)


abkfenris (approved):
Awesome! This is something that I never got to dig into.

We definitely want this for the rest of the site. We might also want to be able to save the datum, and a default time range (say if they select 3 days, rather than a specific date range).

PR #2946, opened 2024-06-17
Fix slow loading PredWL dataset loading

Adelle-Pitsas:
This increases fetch abort timeout to 10000ms to avoid error out on big predWL datasets - per our convo in planning


src/Features/ERDDAP/hooks/tabledap.ts
```diff
@@ -66,6 +66,7 @@ export function urlBuilder(timeSeries: PlatformTimeSeries[], startTime?: Date, e
   const variables = timeSeries.map((ts) => ts.variable)
   const server = timeSeries[0].cors_proxy_url ? erddapService + timeSeries[0].cors_proxy_url : timeSeries[0].server
 
+  console.log("bananas", tabledapUrl(server, timeSeries[0].dataset, variables, constraints))
```
abkfenris, line 69:
```suggestion
```

PR #2987, opened 2024-07-17
make y-axis dynamic for min and max

Adelle-Pitsas:
This finally fixes the y axis for tide gauge data

- On gauges that don't have flood thresholds, it calculates the max value from current timeseries and makes it a soft max
- On gauges that do have flood thresholds, it adjusts the max to be .5 units (metric or english) above that threshold (so we don't have enormous max plotbands--see before pic below)

Before (no thresholds):
Screenshot 2024-07-17 at 4 14 59 PM

After (no thresholds):
Screenshot 2024-07-17 at 4 15 31 PM

Before (thresholds):
Screenshot 2024-07-17 at 4 15 58 PM

After (thresholds):
Screenshot 2024-07-17 at 4 16 21 PM

### DIFF
--- a/src/Features/ERDDAP/waterLevel/chart/largeTimeSeriesChart.tsx
+++ b/src/Features/ERDDAP/waterLevel/chart/largeTimeSeriesChart.tsx
@@ -38,9 +38,9 @@ interface Props {
   /** Name of data */
   name: string | undefined
   /** Soft minimum for Y axis */
-  softMin: number | undefined
+  softMin: number
   /** Soft maximum for Y axis */
-  softMax: any
+  softMax: number
   /** Unit system to display in */
   unitSystem: UnitSystem
   /** Data type to display */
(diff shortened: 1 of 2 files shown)


src/Features/ERDDAP/waterLevel/chart/largeTimeSeriesChart.tsx
```diff
@@ -123,7 +123,11 @@ export function LargeTimeSeriesWaterLevelChart({
         </XAxis>
         <YAxis
           softMin={softMin}
-          softMax={floodThresholds ? floodThresholds?.Major?.minValue + 3 : softMax[unitSystem]}
+          softMax={
+            floodThresholds
+              ? floodThresholds?.Major?.minValue + dataConverter.convertToNumber(0.5, unitSystem)
```
abkfenris, line 128: Do we want to get the max of all the thresholds, in case there isn't Major, or one of the impacts is higher? (often with USGS, I've seen Major being some point that is hit every 10 years or so, and there are much higher impacts).

abkfenris (approved):

src/Features/ERDDAP/waterLevel/chart/largeTimeSeriesChart.tsx
Adelle-Pitsas: Good thinking. I'll refactor to get the highest.

PR #3005, opened 2024-07-26
Adelle/current wl clean up

Adelle-Pitsas:
Synchronizes `usePlatform` and `useDataset` so the explicit "recent value" in the sidebar reflects the most recent reading taken as seen in the graphed timeseries. This was done to avoid extra calls to ERDDAP and used cached data to our advantage.

Logic for combining happens in: `/Features/ERDDAP/hooks/TableDAPComponents.tsx`

This logic was then used for the rest of the app so all most recent values in main part of app (not just in water level portion) are also updated with most accurate value.

Ex.
Screenshot 2024-07-26 at 4 01 09 PM

### DIFF
--- a/src/Features/ERDDAP/hooks/TableDAPComponents.tsx
+++ b/src/Features/ERDDAP/hooks/TableDAPComponents.tsx
@@ -23,7 +23,6 @@ interface UseDatasetsProps {
 
 export interface UseDatasetsRenderProps {
   datasets: DataTimeSeries[]
-  platforms: PlatformTimeSeries
 }
 
 interface UseQueryGroupResult {
@@ -76,7 +75,6 @@ export const UseDatasets: React.FunctionComponent<UseDatasetsProps> = ({
       }
     }
   }
-  let updatedPlatforms
 
   if (loadedDatasets && platformId) {
     const platforms: any = queryClient.getQueryData(["buoybarn-platforms"])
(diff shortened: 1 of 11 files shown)


src/Features/ERDDAP/hooks/TableDAPComponents.tsx
```diff
@@ -72,14 +76,48 @@ export const UseDatasets: React.FunctionComponent<UseDatasetsProps> = ({
...
+          readings: updatedReadings,
+        },
+      }
+
+      updatedPlatforms = {
+        ...platforms,
+        features: platforms.features.map((f) => (f.id === platformId ? updatedPlatform : f)),
+      }
+      // queryClient.setQueryData(['buoybarn-platforms'], platforms);
+    }
```
abkfenris, line 111: How about comparing the reading times, and setting a flag if there is any updated readings or not? 

If there are updated readings, then call `setQueryData` and trigger a re-render. Then during the re-render the times should be the same, so it shouldn't get called again, and thus avoid re-rendering.

src/Features/ERDDAP/hooks/TableDAPComponents.tsx
```diff
@@ -73,6 +76,42 @@ export const UseDatasets: React.FunctionComponent<UseDatasetsProps> = ({
...
+        ...platforms,
+        features: platforms.features.map((f) => (f.id === platformId ? updatedPlatform : f)),
+      }
+      const updatedValues = updatedReadings.map((r) => r.value).toString()
+      const originalValues = platform.properties.readings.map((r) => r.value).toString()
+      if (originalValues !== updatedValues) {
+        queryClient.setQueryData(["buoybarn-platforms"], updatedPlatforms)
+      }
+    }
+  }
```
abkfenris, line 113: It might be worth some extra inline comments to document the intent of this code since it's both doing something cool, and not completely obvious.

abkfenris (approved):

PR #3006, opened 2024-07-26
Fix JSURL import failure

abkfenris:
JSURL was failing to import with `Attempted import error: 'jsurl2' does not contain a default export (imported as 'JSURL').` which caused the modeling pages to fail to load.


abkfenris:
@Adelle-Pitsas This shouldn't cause much of a change, but feel free to land it when it won't cause too much of a hiccup.

PR #3012, opened 2024-07-29
Redirects for tide gauges

### LINKED ISSUE #3011: Redirects for tide gauge signs
We're going to be putting up signs at some of the new tide gauges. Rather than using a URL that may change over time, we're going to short URLs with Next.js's [redirect options](https://nextjs.org/docs/app/building-your-application/routing/redirecting#redirects-in-nextconfigjs).

[Here are the current signs planned](https://gmri.app.box.com/s/yu9dcfp3xac6jno336991tetwhf7utzh)

We need to decide on formats for the redirect separator (`/s/`, `/short/` or `/redirect/` or otherwise), and how the location slugs will be formatted.

We also need to make sure if they are generating QR codes, they don't go through another redirection service along the way (some QR code generators sneak in tracking junk).

abkfenris:
This sets up URLs using Next.js redirect functionality to provide slightly shorter URLs that more importantly are stable (and can be adapted if the naming of the gauges changes, or they should be pointed towards the water level pages instead). 

This should be deployed to the main site before the signs are installed.

Closes #3011

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -13,6 +13,31 @@ const nextConfig = {
   //   disableStaticImages: true
   // }
   output: "standalone",
+  redirects: async () => {
+    return [
+      // Redirects for tide gauge signs https://github.com/gulfofmaine/Neracoos-1-Buoy-App/issues/3011
+      {
+        source: "/s/fore-river-tide-gauge",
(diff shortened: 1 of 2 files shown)


Adelle-Pitsas (approved):
Looks good!

Issue #3015, opened 2024-07-30
Loading states


Adelle-Pitsas:
Places where Next is a little slow, add loading states for better UX

Adelle-Pitsas:
This has been closed by #3139

PR #3043, opened 2024-08-08
Adelle/loading states

Adelle-Pitsas:
This adds a generic loading state to places where next is a little slow:

https://github.com/user-attachments/assets/31add366-f153-4a8f-8b42-c71f2195aa6b

Addresses #3015


abkfenris:
Hmm. That feels a bit too in your face, and blocks from other interaction, and causes the map to flash in and out. 

What about animating the existing logo in the menu bar, and pushing down `loading.tsx` into the different @ components, so it's a subset of the page that shows a loading state rather than the whole thing getting blanked?

PR #3060, opened 2024-08-14
Adelle/obs pred indication

Adelle-Pitsas:
This finally fixes the RLayers issue with not rendering the most recent/updated data being passed to the map component.

Fixes:
- Adds `key` prop to `rLayers` point so with any new rerenders, it creates a completely new component (and uses newest/recent color to reflect tides passing data)

New functionality:
- Inscribed circular point indicates flood threshold surpassed in last 24 hours. 
- Square encompassing each point indicates predicted flood threshold surpassed in next 24 hours

If we want to include this new functionality for upcoming deployment, we'll need to add a few things to the legend to symbolize the obs vs pred

Screenshot 2024-08-14 at 1 41 09 PM


abkfenris:
Very cool! I think this is probably a bigger change than we want to try to get in today, especially before I have a better understanding on how the request pattern will change, and if we should do any work on Buoy Barn to help make sure it can be shown efficiently.

PR #3061, opened 2024-08-14
Wind Speed in Knots

abkfenris:
UConn's ERDDAP server is only reporting wind speed in knots and is unlikely to change, so adding a new data type that will be able to accept their data.


Adelle-Pitsas (approved):
Speedy!

abkfenris:
@Adelle-Pitsas It's not a huge change and unlikely to break other things, plus I knew were to start tweaking.

Issue #3083, opened 2024-09-04
Slow loading with Next.js


abkfenris:
It looks like React-Query have updated some of their suggestions on how to play with the Next.js app router: https://tanstack.com/query/latest/docs/framework/react/guides/advanced-ssr#alternative-use-a-single-queryclient-for-prefetching

There are probably some places we can optimize and pre-fetch better, and others where we can get the right loading transition to give better feedback.
```[tasklist]
### Tasks
- [ ] Perfomance monitoring
- [ ] https://github.com/gulfofmaine/Neracoos-1-Buoy-App/pull/3103
- [ ] Hack large chunks out to get a rough idea where slowness is happening
```

Adelle-Pitsas:
Working on this. I think generally there are a lot of places we are using `use client` where we should be using server components that would prevent us from using the single query client for prefetching approach. Refactoring them to server components has proven to be a bit of a bear. One issue we might be facing is frequently cold starting, where pages take significantly longer to load if a user hasn't hit that endpoint within the recent past (from what I've read 5ish mins). 

For the time being, loading transitions seem to be a good bandaid, but it'd be good to figure out an effective pre-fetch strategy to follow in the future. Also found some folks who may be struggling with similar issues....

https://www.reddit.com/r/nextjs/comments/1at62b5/slow_server_response_on_first_load_using_ssg/
https://www.reddit.com/r/nextjs/comments/zhrx41/waiting_for_server_response_is_taking_too_much/

abkfenris:
I may have linked to the wrong part of react-query doc. What if we start with a top level prefetch of platforms that is cached longer on the server, then have a shorter cache time on the client?

Adelle-Pitsas:
What's been tried:
- Moving components to server side rendering--removing "use client" (Adelle)
  - Doesn't make anything any faster and also is not compatible with our built infrastructure
- Adding `loading.ts` to slow loading routes (Adelle)
  - Doesn't improve routing loading time (the main issue), but is probably a good practice for slower loading components in the future
- Refactor to `getServerSideProps` (Adelle)
  - Isn't compatible with how we are fetching our data (we aren't always fetching in routes)
- Added `Suspense` boundaries (Adelle)
  - This was at the beginning of this journey when I thought the data was the cause of the slow loading. Did nothing and didn't use the fallback component, mainly because this is mostly a routing issue.

abkfenris:
[...]

PR #3103, opened 2024-09-25
Adelle/load time

Adelle-Pitsas:
This is an attempt to fix the long loading times for non-nested routes in next

- Adds `prefetch={false}` to `/models` and `/water-level` routes so the load is less when initially navigating to app
- Adds a `loading.tsx` file up and down the tree to catch all loading components - this is to avoid next thinking it needs to render the entire page before it gets to route. 

Make no mistake... the routing still takes too long for my liking (especially on the first navigation home and `/region`, but the loading components cut down on what seems like the routing time and gives _slightly_ better UI (maybe).

The loading component so you don't have to fidget with it to get to see it:
https://github.com/user-attachments/assets/021744c0-bb97-4ec5-a536-bbf8f3cb24c6


Adelle-Pitsas:
This has mostly been resolved by #3083

PR #3139, opened 2024-09-27
Global loading status

abkfenris:
This uses https://github.com/Skyleen77/next-nprogress-bar to put a loading status line across the top (in buoy yellow) during transitions between pages. 

The native Next.js way tends to hide parts of the page when they are being replaced, so this is a different way of showing that status change.


abkfenris:
We may want to update the styling of the loading alerts to match.

Adelle-Pitsas:
Added the spinner to the loading component and also added a quick function to get the name of the dataset being fetched for some context for the user (especially in the case of overlayed datasets on a graph where one has loaded and the other just shows a generic "Loading Data" message that could be misleading).

Water level page (exemplifying overlay datasets and why naming the loading dataset might be helpful):
Uploading Screen Recording 2024-10-10 at 12.08.20 PM.mov…

Platform page:
https://github.com/user-attachments/assets/fa49f7c2-b182-4c65-9662-91a4f1119496

If we don't like it, I'm happy to edit/remove it!

abkfenris:
Ohh, including the dataset info is very nice. I wonder about coloring the spinner buoy yellow to connect color wise with the loading bar?

abkfenris:
I went and standardized loading alerts to have spinners, made them light colored backgrounds with yellow spinners to match the loading bar.

https://github.com/user-attachments/assets/bc664d18-b7c1-4c0f-954f-39289e216b44



I also consolidated warning alerts into a single spot, and may have fixed some routing funkiness with the map (acted as a regular link when clicking the popup but used the Next/loading router when clicking the dot previously, now should use the router for both, but still present a link as a popup).

abkfenris:
Gah, A01 is current failing QC checks causing Playwright to fail.

PR #3194, opened 2024-10-23
Adelle/site timeframe selector

Adelle-Pitsas:
This PR puts up a timeframe selector for observations in main MD site. It uses the already created url search params pattern and component used in water level section of site.
 

https://github.com/user-attachments/assets/3aa30b2e-ef5c-4146-ac38-ed7d2eb2b947

Happy to receive feedback--a few more things that I know need to be done on this:
- [x] "Revert to default date" button works weirdly--will have to adjust so it doesn't change water level and works well for other obs
- [x] - Styling needs some work
- [x] - Need to adjust how much time user is able to select. For water level we don't allow more than two weeks of data from server, but here we might want more (a month was floated as the limit for how much data to be fetched in one go)
- [x] Fix Wind links


abkfenris:
It kind of sits awkwardly in it's own space right now, so I wonder if there is a way to bring it more inline. It also would become more hidden for data types that show multiple values such as water temperature https://mariners.neracoos.org/platform/A01/observations/sea_water_temperature


It may help to think about styling together with chart (stacked or overlaid for multiple charts) styling and tabular display and export.

Adelle-Pitsas:
Yeah, I think having it at the top would be better ux. Thoughts on this layout? Definitely have to spruce up the functionality for something with multiple graphs like sea water temp--gets a little funky with url searchParams
Screenshot 2024-10-23 at 3 39 18 PM

abkfenris:
Ya, that looks much cleaner. The URL should be fine as is for multiple as it's based on standard name. We'd probably have an additional search param for stacked/chart...

abkfenris:
I bet we could allow a year (especially if we're trying to replace some graphing and download functionality, which reminds me that for a single dataset we already have the kinda hidden but helpful links to ERDDAP...). I think a year's worth from most platforms comes to about a 1 MB JSON response. 

I tried ten years just because, and that got a funky and made Firefox ponder it's role in the universe.

image

Adelle-Pitsas:
Okay, I think this is good to go but throw in some tasks if you think there is something I missed or needs to be changed:
- Increases timeframe selection to 1 year
- Styles for mobile (see below)
- Functional "revert" button
- Addresses the stacked charts to only show 1 time selector


https://github.com/user-attachments/assets/f52e585c-d5fc-46aa-8d40-1b09e37d51be


https://github.com/user-attachments/assets/e178e2c6-d77e-4c85-a52e-a9a617fdb347

abkfenris:
I think I would have it always visible for mobile. The button isn't quite self explanatory. Maybe the [`calendar-days`](https://fontawesome.com/icons/calendar-days?f=classic&s=regular) icon, and some text if we're going to keep it. I'd also explore moving the links from the popover to also live there (and check that their date ranges are updated with the chart).

Adelle-Pitsas:
Sounds good. Do you want those links still in popover form? Or have them more visible?

abkfenris:
If there is a good way to integrate them so they are more visible, we can remove the popover.

Adelle-Pitsas:
Okay, Alex and I talked and there isn't a good way to change the "i" since its connected to specific graphs (and won't work with stacked graphs) so tabling moving those links for now, but
- Changed the text on the button in the mobile view (I think its good mobile UI to hide it within the button since it takes up decent amount of room)
- Added graphing url to the data access popover

src/Features/ERDDAP/Platform/Observations/Condition/index.tsx
```diff
@@ -46,8 +57,28 @@ export const ErddapObservedCondition: React.FunctionComponent<Props> = ({ platfo
...
-            {({ dataset }) => <ChartTimeSeriesDisplay {...{ dataset, standardName, unitSystem }} timeSeries={ts} />}
+          {index === 0 && (
+            <div>
+              <Button color="light" outline onClick={() => toggle()} className="timeframe-collapse-button">
+                <p style={{ marginBottom: 0, marginRight: "5px", color: "#083d52" }}>Change Timeframe</p>
+                <Calendar width={"20px"} height={"20px"} />
+              </Button>
+              <Collapse isOpen={isOpen} className="timeframe-collapse">
+                <div className="observation-timeframe-selection">
+                  <TimeframeSelector graphFuture={false} />
```
abkfenris, line 68: Maybe the `TimeframeSelector` is an is part of the `<h4>` (or really a parent component of it), rather than trying to be manipulated with CSS to play nice at all sizes?

abkfenris (commented):
I'm not sure we need a toggle at this size, and the selector ends up overlaying the latest data on some charts, which is probably what folks are most interested in 

image

At some other sizes it also ends up overlaying the variable title h4.

image

Wind on at least a few different platforms isn't loading. 
image

src/Features/ERDDAP/Platform/Observations/Condition/index.tsx
abkfenris: It happens when navigating from the current conditions page via the card headers for any of those conditions, and also if you click the 'last values' between the station info and the unit system. 

https://github.com/user-attachments/assets/6f3ff55f-dfd1-4ba5-aff9-fa81672f2960

I think the issue is that it's getting nulls or empty strings when start and end params aren't set, so we may want to add fallback values https://github.com/gulfofmaine/Neracoos-1-Buoy-App/pull/3194/files#diff-b0b4393f4c4f58be180f325dd4511e323a5fa8ca0522da583515f424a0629467R40-R41

Let's not add params to links by default and have them bare. https://github.com/gulfofmaine/Neracoos-1-Buoy-App/pull/3194/files#diff-0301babb06bd3617e76d92b6de2baf63f786c2a63259bd8ce98600660a4b9e6bR78-R116

Looks and behaves better, but maybe a touch of padding/margin (I can't ever remember which is which side).

image

Adelle-Pitsas: The most recent commits should take care of
- The margin between the title and the timeframe selector (and aligns to the center so it doesn't look too terrible)
- Default time params so we don't change the link path, but set default values for start and end, and only when the user changes the timeframe should the params in the url change. 

Thanks for the feedback on all of this! Keep it comin!

src/Features/ERDDAP/hooks/TableDAPComponents.tsx
```diff
@@ -157,10 +158,11 @@ export const UseDataset: React.FunctionComponent<UseDatasetProps> = ({
...
   startTime,
+  endTime,
   loading,
   error,
 }) => {
-  const { isLoading, data } = useDataset(timeSeries, startTime)
+  const { isLoading, data } = useDataset(timeSeries, startTime, endTime)
 
   if (isLoading) {
     if (loading) {
```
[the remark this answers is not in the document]
Adelle-Pitsas: I'll look into this!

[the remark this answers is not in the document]
Adelle-Pitsas: Okay, so we weren't actually showing `PlatformLoadingAlert` on non-wind observations (i.e. fetching one dataset using `useDataset`) so that should be fixed now. 
I'm not sure why wind is not working locally on your end as it is set up to use `useDatasets` and the fallback for loading is showing the loading alert. I can see it always locally. I'm wondering if you aren't seeing the loading components because its loading quickly? If you adjusted the throttling in the dev tool network tab, are you able to reproduce what you were seeing previously (no loading components)? If its still not working, I can dig deeper but from what I can see, wind should be showing loading components for all three of the datasets its loading!

Here's it on slow 4g:
https://github.com/user-attachments/assets/171c3598-340b-4dfa-83d4-2838cd6a1c87

Issue #2860, opened 2024-04-18
Watel level UI enhancements


Adelle-Pitsas:
These next few are enhancements not bugs...

New features/behaviors:
- [x] Add vertical line to show current time on plot

Changes to current features/behavior (open to discussion)

- [x] Default time is GMT - do we want to convert to EST?
<img width="958" alt="timestamp_GMT" src="https://github.com/gulfofmaine/Neracoos-1-Buoy-App/assets/3279645/917ea4f4-91f5-4bf5-b5c5-b40cc20679da">

- [ ] For stations with not a lot of difference in benchmarks, it's hard to see the difference in the two levels. the https://water.noaa.gov has this problem too, so it may be a tough one to solve. 
<img width="943" alt="Benchmarks_too_close" src="https://github.com/gulfofmaine/Neracoos-1-Buoy-App/assets/3279645/02ac184c-f3c9-444b-8b62-348d314f23b8">

- [x] For stations that have a lot of additional data do we want to limit what we show in the station info? Noticed this on a station that had 6-7 variables.

<img width="445" alt="Screenshot 2024-04-12 at 5 13 43 PM" src="https://github.com/gulfofmaine/Neracoos-1-Buoy-App/assets/3279645/783b1f03-3034-410e-8665-73d336e51267">

_Originally posted by @youngmorse in https://github.com/gulfofmaine/Neracoos-1-Buoy-App/issues/2701#issuecomment-2052542150_

Adelle-Pitsas:
- [ ] Add fixed height and scroll to sensor dropdown in sidebar

Adelle-Pitsas:
- [ ] refactor "Feet" to "ft" in in side bar to account for "1 Feet"

abkfenris:
How about we just standardize distance measurements to Smoots?

youngmorse:
Adding a few new issues/requests that I ran into testing the Next.js viewer in production (with enough resources to be fast!)

- [ ] Alpha sort the station list

![Screenshot 2024-09-27 at 1 16 40 PM](https://github.com/user-attachments/assets/79726e36-b5b1-4719-b90d-6c1d365cefa3)

- [ ] Even with alpha sort, the station IDs are mysterious and hard to decipher unless you know them. We may want to consider using Station Name (and maybe ID in parentheses) from the drop down. And use smaller font. 

[...]

PR #3198, opened 2024-10-24
seperate sensors by region

Adelle-Pitsas:
This addresses a request from Tom/Anna to have the sensors separated by region in water level list page (similar to how platforms are separated in main site) #2860 

screencapture-localhost-3000-water-level-2024-10-24-15_44_45


Adelle-Pitsas:
A go from Alex via slack

PR #3213, opened 2024-11-07
Adelle/default search params

Adelle-Pitsas:
This PR:
- Fixes the outline for the geolocated circles in the map
- Updates searchParams so when user goes to `sensor/sensorID` the default search params are 2 days previous, 3 days in the future, mllw. @abkfenris I'd love it if you could try and break this. There were some edge cases I found last night when I was testing it and fixed them but I'm not confident I got all of them. Any place where the user can be routed to the sensor page I'd love to have checked. The user flow should go as follows:
  - When a user selects a sensor from the water sensor list page (`/water-level`) they should be brought to the sensorID page with no search params in url, but the page should default to `2 days prior`, `3 days future`, `mllw`.
  - When a user selects either a different datum or different time selection, the url should update _only_ with that params updates (i.e. if a user selects a new datum, the datum params shows in the url, but the date based params stays the same and are not updated)
  - When a user selects a NEW sensor from the map or sensor selector in the sidebar, whatever search params are in the url should stay the same but the sensor should change (obviously). 
  - When the user CLEARS the timeframe (in the timeframe selection box) the time based search params should be removed, and graph and timeframe should revert to default but if there are datum search params in url, those should stay the same. 

I think this is the most intuitive url pattern in that the url changes based only on what the user interacts with and the search params don't reset based on selecting a new sensor. Happy to chat this through if needed.


src/Features/ERDDAP/Map/WLPlatformLayer.tsx
```diff
@@ -133,13 +141,15 @@ const Layer = ({ platform, url, router, radius, color, floodThreshold }) => {
   useEffect(() => {
     setKey((prevKey) => prevKey + 1) // Increment key on state change
   }, [color])
+
+  console.log("bananas", color)
```
abkfenris, line 145:
```suggestion

```

src/Features/ERDDAP/Map/WLPlatformLayer.tsx
```diff
@@ -94,29 +94,37 @@ export const WLPlatformLayer = ({ platform, selected, old = false }: PlatformLay
...
-        searchParams.get("datum") as DatumOffsetOptions,
-      )
-    : buildSearchParamsQuery(getIsoForPicker(daysAgoRounded(2)), getIsoForPicker(daysInFuture(3)), "datum_mllw_meters")
+  // const query = isSensorPage
+  //   ? buildSearchParamsQuery(
+  //       searchParams.get("start") as string,
+  //       searchParams.get("end") as string,
+  //       searchParams.get("datum") as DatumOffsetOptions
+  //     )
+  //   : ""
```
abkfenris, line 109:
```suggestion
```

src/Features/ERDDAP/waterLevel/timeframeSelector.tsx
```diff
@@ -16,8 +23,8 @@ export const TimeframeSelector = ({ graphFuture }: { graphFuture: boolean }) =>
   const params = useParams()
   const pathname = usePathname()
   const searchParams = useSearchParams()
-  const [startTime, setStartTime] = useState<any>(searchParams.get("start"))
-  const [endTime, setEndTime] = useState<any>(searchParams.get("end"))
+  const [startTime, setStartTime] = useState<any>(searchParams.get("start") || getIsoForPicker(daysAgoRounded(2)))
+  const [endTime, setEndTime] = useState<any>(searchParams.get("end") || getIsoForPicker(daysInFuture(3)))
```
abkfenris, line 27: Seeing this pattern duplicated but slightly different makes me think we might want to refactor these out into self contained hooks that capture the functionality and complexity of `useSearchParams`, `useState`, and `useEffect`, so we might just end up with `useStartDate`, `useEndDate`, and `useDatum` that are more self explanatory (and may help contain if we refactor datums into using redux/session store).

abkfenris (approved):
So far I haven't been able to technically break anything. 

The one thing that feels off now is that we persist a users choice of units, but not datums between fresh loads with no params specified. I haven't fully thought through how that interplay between redux storage and sharing links might work though, so that might be something to ponder for later.

src/Features/ERDDAP/waterLevel/timeframeSelector.tsx
Adelle-Pitsas: Cool, yeah, I can wrap that into its own hook. And agreed about the unit selection. It might be we get some feedback from users soon on what they are expecting and how we can engineer it to be more intuitive.

Adelle-Pitsas: I'm going to do this in another branch and merge this one

PR #3215, opened 2024-11-08
expose water level in nav bar on prod

Adelle-Pitsas:
This exposes the `Water Level` nav item on the nav bar in production


abkfenris (approved):
🥳

PR #3229, opened 2024-11-18
Adelle/fix invisible flood thresholds

Adelle-Pitsas:
This fixes a few things that I saw were missing last week:
- We lost flood thresholds when I shortened the url, so this adds them back with the default datum
- Tide gauges without ability to convert datums still had the long url with all the search params as a default link--this removes those search params as a default and switches the the new short url pattern.
   - This additionally clears the datum search param when: a user changes the datum on a platform that has datum conversation and then clicks on a gauge that does not have datum conversions (previously it was keeping the datum search params in the url, but was still showing data for navd88 which is misleading.)


abkfenris:
I think these may be loading timing errors in tests.

PR #3264, opened 2024-12-04
Adelle/forecasts on sensor page

### LINKED ISSUE #3259: Incorporate forecast data into sensor page
Incorporating forecasting data into sensor page:

- [x] Take the highest value of either forecast or predicted and display as the square on the map
- [x] Add forecast timeseries to the graph
- [x] Add tabs on the tide table for predicted, and any forecasts coming in (potential for more than 1 forecast come in)

Adelle-Pitsas:
This PR incorporates Forecasts into the water sensor page and closes #3259 
- Takes the larger value to display as the future flooding indicator on the map
- Updates `useDatasets` to get datasets by type instead of index so we don't mistakenly show a forecast as a prediction
- Adds Forecasts to graph
- Adds variable tabs on tides table
- All functionality written with the potential for more than one forecast
(sorry I had to add bootstrap as a dependency... tabs for literally everything else wasn't working with our page structure)

Screenshot 2024-12-04 at 2 31 13 PM


package.json
```diff
@@ -28,6 +28,7 @@
     "query-string": "^9.1.1",
     "react": "^18.3.1",
     "react-app-polyfill": "^3.0.0",
+    "react-bootstrap": "^2.10.6",
```
abkfenris, line 31: Hmm, we already had bootstrap, and tabs here https://github.com/gulfofmaine/Neracoos-1-Buoy-App/blob/main/app/(platformMap)/%40bottom/platform/%5BplatformId%5D/tabs.tsx using reactstrap

image

src/Features/ERDDAP/hooks/tabledap.ts
```diff
@@ -149,6 +149,8 @@ export const getDatasetGroup = async (fetchGroup: FetchGroup, startTime?: Date,
   const resultTimeSeries = resultToTimeseries(
     json,
     fetchGroup.datasets.map((ts) => ts.variable),
+    ...fetchGroup.datasets.map((ts) => ts.dataset_public_name),
```
abkfenris, line 152: Will this be predictable when not all timeseries will have a `.dataset_public_name` that isn't null?

src/Shared/colors.ts
```diff
@@ -4,6 +4,7 @@ export const colors = {
   coastalMeadow: "#2ab78f",
   buoyYellow: "#ffcd22",
   whatOrange: "#ffae6d",
+  brighterOrange: "#fc8042",
```
abkfenris, line 7: Is this from the NERACOOS styles?

src/Features/ERDDAP/hooks/tabledap.ts
Adelle-Pitsas: I can't wrap my brain around the double negatives hah. Can you say this another way?

abkfenris: I see what is happening here now. You're supplying the function with a map of a couple of different things in the same order, and then it extracts and uses the `dataset_public_name` if avaliable.

Adelle-Pitsas: Yes, that's whats going on. It should account for null values in `dataset_public_name`

src/Shared/colors.ts
Adelle-Pitsas: No its not from NERACOOS styles, but to me, it seems like none of the current colors have enough of a contrast to be effective as a third color in the graph:
This is what it looks like with neracoos' `whatOrange`:
Screenshot 2024-12-05 at 12 59 24 PM
The best contrast for the colors we have is `neracoosBlue` but that just looks like a bolder version of the "Observed" timeseries. 
Screenshot 2024-12-05 at 1 00 20 PM

The `brighterOrange` that I added looks like this:
Screenshot 2024-12-05 at 1 01 24 PM

Happy to change it to whatever you think is best! Just want it to be clear what timeseries is which (especially for folks who might be visually impaired)

package.json
Adelle-Pitsas: Okay I got it working with reactstrap--just had to create a new reusable component that doesn't change the url and only changes the tab panels below the tabs.

src/Features/ERDDAP/waterLevel/TidesTable.tsx
```diff
@@ -1,11 +1,32 @@
...
+  NavLink,
+  Row,
+  TabContent,
+  Table,
+  TabPane,
+} from "reactstrap"
 import { WATER_LEVEL_STANDARDS } from "Shared/constants/standards"
 import { round } from "Shared/math"
+// import Tab from "react-bootstrap/Tab"
+// import Tabs from "react-bootstrap/Tabs"
```
abkfenris, line 23:
```suggestion
```

src/Shared/colors.ts
abkfenris: Is it cycling through the colors or are you picking them? 

What about using different dash styles for predicted vs forecast? That's what we are doing on the platform forecasts https://github.com/gulfofmaine/Neracoos-1-Buoy-App/blob/b90c86d5ea4056824cb3ea2889b6139aaf296f75/src/Features/ERDDAP/Platform/Forecasts/Page/index.tsx#L105

image


https://mariners.neracoos.org/platform/E01/forecast/air_temperature Though now I'm thinking we should probably make the observed solid and the forecasts dashed there

Hohonu uses dashes for their forecasts and dots for their observations

image

CO-OPs is solid for both

image

We show prelim (but out of date because I gonna fix something with it...) data dashed on the Climate dashboard 
image


- https://www.highcharts.com/demo/highcharts/line-forecast
- https://api.highcharts.com/highcharts/series.line.dashStyle
- https://api.highcharts.com/class-reference/Highcharts#.DashStyleValue

abkfenris (approved):

PR #3921, opened 2026-01-29
Clean-up-dev-environment

abkfenris:
Went through the readme to hopefully make things a little clearer. I wanted to make sure the test commands matched up with what was documented and what was getting run in Github Actions.


shindelr:
Looked over the Readme. Definitely the Make and NPM commands line up better now, thanks for making those changes! I haven't tried running the test commands yet, but will as soon as we get rolling on the project.

Issue #3875, opened 2026-01-08
MD Refresh - Style and Grids


youngmorse:
Overall style refresh includes updates to Bootstrap grid, CSS changes for color/font/icons, improved layout for station detail page

Bootstrap Grid layout Desktop and Mobile

- [ ] Color update #3987 
- [ ] Typography #3988 
- [ ] Icons #3989 
- [ ] Station Detail cards/colors #3990 

Global

- [ ] Abbreviate units everywhere #3991 
- [ ] Additional color styles and icons to reflect updated features #3992 

Station Detail Page style changes

- [ ] Dig provided three new layouts for the station detail page. We like the middle option, it includes background color behind map/attribution and new format for latest data using cards/tiles by variable.  (See figure below.) #3993 

**Bootstrap Grids & Styles**
<img width="1148" height="899" alt="Image" src="https://github.com/user-attachments/assets/43003808-f268-46ef-b52a-caf674c046a4" />

**Colors & Typography**
<img width="1804" height="1617" alt="Image" src="https://github.com/user-attachments/assets/95016017-0743-4151-b63a-29b3d24facfc" />

---
**Station Detail Page style changes**
<img width="2107" height="722" alt="Image" src="https://github.com/user-attachments/assets/185916c3-6ac1-465e-8d32-2ca4a6decf57" />

tshyka:
I think you should do your best to imitate the middle option because it was the one we chose.

shindelr:
Sounds good!

PR #4060, opened 2026-04-07
Modified display unit default

shindelr:
@cgalvarino @abkfenris 

Groundwork infrastructure to begin working on #3875 "Abbreviate units everywhere". Chatting with Alex, the preferred move here is to keep the ability to display long unit names rather than a straight replacement. 

Instead of building a new abbreviation function and trying to make the change from displayName across the entire codebase, I just changed the functionality to display the abbreviated unit with a fallback to the original choice if unavailable. Made a small getter function displayFullUnitName to allow us to directly snag that long name in a future tooltip component.


shindelr:
@abkfenris I think this is pretty much good to go if it matches your vision for the future. The failing codacity check is about the class property names not being in camel case.

abkfenris (approved):

PR #4028, opened 2026-03-19
groundwork for getting top 2 obs inside map hover

cgalvarino:
#3876

### NOT READY FOR MERGE

@shindelr 

Reusing as much as possible from standard latest obs workflow to get the top 2 obs inside a mouseover.  Current status?  Fugly but functional.

image

**Updated Station Hover Popup Spec**
- Background color: Black
- Text color: White
- Include top 2 conditions
- Include last updated date and time
- Text: ‘Paragraph Bold’ for station name; ’Caption Bold’ for data title; ’Caption’ for data units + Last Updated

Issue #3876, opened 2026-01-08
MD Refresh - Landing Page


youngmorse:
**Landing Page - [https://mariners.neracoos.org/](https://mariners.neracoos.org/)**

Updates to the landing page are a mix of style/font modifications, minor content/layout changes, improved functionality for the map, and a change from the 'Superlatives' (e.g., highest waves, highest winds to Nearby Stations by user location).  The Nearby Stations recommendations is likely a Phase II effort since we don't have user location  on the current app. 

Figma mockup images attached w/ updates outlined in text below. Before image at the bottom.

**New Desktop**
<img width="1334" height="2579" alt="Image" src="https://github.com/user-attachments/assets/1b1cbe1d-c37d-4ad2-a50f-402837c57b96" />

**Map Key**

- Added key to be anchored in top left of map
- Text: ‘Caption’; Black 65
- Background color: White
- Border: 1px, outside, Black 20
- Border Radius: 2px
- Padding: 8px 12px 8px 12px

**Updated Station Hover Popup**

- Background color: Black
- Text color: White
- Include top 2 conditions
- Include last updated date and time
- Text: ‘Paragraph Bold’ for station name; ’Caption Bold’ for data title; ’Caption’ for data units + Last Updated

**Icon Buttons**

- Added ‘zoom to current location’ (top)
  - Icon: ‘navigation-variant’, Black 65
- Added ‘expand map’ (middle)
  - Icon: ‘arrow-expand’, Black 65
- Background color: White
- Border: 1px, outside, Black 20
- Border Radius: 2px

**Logo & Subtitle**

- 16 px space to right of logo
- ‘Paragraph’ style; White
- Link logo and text to Landing page

**Map Height**

- Map height should match the height of the content on the right side
- On Landing, map height minimum should be 590px

**Nav Links**

- Removed ‘Home’
- Renamed ‘Regions’ to ‘Station List’
- Inactive link color: White 55
- Active link color: White

**Get Started**

- Renamed title to ‘How to Use Dashboard’ (H3)
- Background color: Black 5
- Title & Paragraph color: Black
- Bullets: ‘Paragraph
- Updated Copy

**Latest Conditions**

- Renamed title to ‘Top Wind & Waves—All Regions’ (H3)
- Moved to the right of the map
- Added info icon popup to house Last Updated timestamp and other information
- Background color: Black 5
[...]

PR #4061, opened 2026-04-08
Top 2 observations in map hover

shindelr:
@cgalvarino @abkfenris 

Addresses: #3876
**Updated Station Hover Popup**
- Background color: Black
- Text color: White
- Include top 2 conditions
- Include last updated date and time
- Text: ‘Paragraph Bold’ for station name; ’Caption Bold’ for data title; ’Caption’ for data units + Last Updated

Building off of Charlton's PR: https://github.com/gulfofmaine/Neracoos-1-Buoy-App/pull/4028

**Picked up at**
Screenshot 2026-04-08 at 12 13 46 PM

**After**
Screenshot 2026-04-08 at 12 11 19 PM

Messed around for a while trying to get the styling correct with the original approach and found that the coupling to the sidebar table was too tight. Changes to the map hover --> changes to the sidebar

**Notes:**
- Added two components: `MapItemPopup` and `LimitedItems`
- Pulled "last updated" logic out into its own helper for readability: `LastUpdated` 
- Moved the short name threshold into `MapItemPopup`
- Reduced BS card complexity in TSX

Suggestions on this approach? I tried to strike a happy medium between keeping the overarching OBS workflow and allowing the map hover to be its own quite different UI component.

### DIFF
--- a/src/Features/ERDDAP/Platform/Observations/Table/item.tsx
+++ b/src/Features/ERDDAP/Platform/Observations/Table/item.tsx
@@ -22,7 +22,6 @@ interface TableItemProps {
   platform: PlatformFeature
   timeSeries: PlatformTimeSeries
   unitSystem: UnitSystem
-  useShortNameThreshold?: number
 }
 
 type TableItemDisplayProps = Pick<TableItemProps, "timeSeries" | "unitSystem"> &
@@ -48,14 +47,10 @@ const TableItemDisplay: React.FC<TableItemDisplayProps> = ({
   )
 }
 
-export const TableItem = ({ timeSeries, unitSystem, platform, useShortNameThreshold }: TableItemProps) => {
+export const TableItem = ({ timeSeries, unitSystem, platform }: TableItemProps) => {
   const tooltipId = `${timeSeries.data_type.standard_name}-tooltip`
 
   let name = timeSeries.data_type.long_name
-  if (typeof useShortNameThreshold !== 'undefined') {
-    name = (name.length > useShortNameThreshold ? timeSeries.data_type.short_name : name) || name;
-  }
-
   if (timeSeries.depth && timeSeries.depth > 0) {
     name = `${name} @ ${timeSeries.depth}m`
   }
@@ -89,5 +84,31 @@ export const TableItem = ({ timeSeries, unitSystem, platform, useShortNameThresh
         </span>
       </OverlayTrigger>
     </Link>
+  )
+}
+
+/**
+ *  Render observation items in a format acceptable for the map hover.
+ */
+export const MapItemPopup = ({ timeSeries, unitSystem }: TableItemProps) => {
+  const shortNameThreshold: number = 50
+  let name =
+    timeSeries.data_type.long_name.length > shortNameThreshold
+      ? timeSeries.data_type.short_name
+      : timeSeries.data_type.long_name
+
+  if (timeSeries.depth && timeSeries.depth > 0) {
+    name = `${name} @ ${timeSeries.depth}m`
+  }
+
+  const unit_converter = converter(timeSeries.data_type.standard_name)
+  const value = unit_converter.convertTo(timeSeries.value as number, unitSystem)
+
+  return (
+    <div className="caption d-flex flex-row gap-1">
+      <strong>{name}:</strong>
+      <span>{typeof value === "number" ? round(value as number, 1) : value}</span>
+      <span>{unit_converter.displayName(unitSystem)}</span>
+    </div>
   )
 }
--- a/src/Features/ERDDAP/Platform/Observations/Table/item.tsx
+++ b/src/Features/ERDDAP/Platform/Observations/Table/item.tsx
@@ -22,6 +22,7 @@ interface TableItemProps {
   platform: PlatformFeature
   timeSeries: PlatformTimeSeries
   unitSystem: UnitSystem
+  useShortNameThreshold?: number
 }
 
 type TableItemDisplayProps = Pick<TableItemProps, "timeSeries" | "unitSystem"> &
@@ -47,10 +48,14 @@ const TableItemDisplay: React.FC<TableItemDisplayProps> = ({
   )
 }
 
-export const TableItem = ({ timeSeries, unitSystem, platform }: TableItemProps) => {
+export const TableItem = ({ timeSeries, unitSystem, platform, useShortNameThreshold }: TableItemProps) => {
   const tooltipId = `${timeSeries.data_type.standard_name}-tooltip`
 
   let name = timeSeries.data_type.long_name
+  if (typeof useShortNameThreshold !== 'undefined') {
+    name = (name.length > useShortNameThreshold ? timeSeries.data_type.short_name : name) || name;
+  }
+
   if (timeSeries.depth && timeSeries.depth > 0) {
     name = `${name} @ ${timeSeries.depth}m`
   }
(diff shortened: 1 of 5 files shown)


src/Features/ERDDAP/Platform/Observations/Table/item.tsx
```diff
@@ -86,3 +86,29 @@ export const TableItem = ({ timeSeries, unitSystem, platform }: TableItemProps)
...
+    timeSeries.data_type.long_name.length > shortNameThreshold
+      ? timeSeries.data_type.short_name
+      : timeSeries.data_type.long_name
+
+  if (timeSeries.depth && timeSeries.depth > 0) {
+    name = `${name} @ ${timeSeries.depth}m`
+  }
+
+  const unit_converter = converter(timeSeries.data_type.standard_name)
+  const value = unit_converter.convertTo(timeSeries.value as number, unitSystem)
```
shindelr, line 105: Not DRY: 104-105 appear in TableItemDisplay and 100-102 appear in TableItem

src/Features/ERDDAP/Platform/Observations/Table/table.tsx
```diff
@@ -63,3 +80,46 @@ export const ErddapObservationTable: React.FC<Props> = ({
...
+        <MapItemPopup key={index} timeSeries={timeSeries} unitSystem={unitSystem} platform={platform} />
+      ))}
+    </div>
+  )
+}
+
+/**
+ * Render a key followed by the last time in a series of dates.
+ */
+const LastUpdated = ({ label, times, className, boldKey }: LastUpdatedProps) => {
```
shindelr: Added `label` and `boldKey` to allow the sidebar to have a different format from the map popup

cgalvarino:
IMHO it's too chunky.  This clobbering is an artifact from the previous implementation, but it's the pits.  Perhaps that's something we jot down as something to fix later.

image

In the meantime, I suppose your changes are sticking to this, warts and all?

Updated Station Hover Popup

Background color: Black
Text color: White
Include top 2 conditions
Include last updated date and time
Text: ‘Paragraph Bold’ for station name; ’Caption Bold’ for data title; ’Caption’ for data units + Last Updated

src/Features/ERDDAP/Platform/Observations/Table/item.tsx
cgalvarino: Not sure what this means.

shindelr: Well, I'm using the same code in two places, figured that wasn't very good style! Could make them into a function but wasn't sure if wanted to add extra complexity for just a few lines?

shindelr:
Paragraph bold with the updated typography indeed looks pretty chunky. Everything's to spec as it is now though

src/Features/ERDDAP/Platform/Observations/Table/table.tsx
```diff
@@ -28,28 +87,31 @@ export const ErddapObservationTable: React.FC<Props> = ({
...
-      ) : (
-        <ListGroup.Item style={itemStyle}>There is no recent data from {platformName(platform)}</ListGroup.Item>
-      )}
+    <ListGroup className="pt-4" as="ul">
+      <ListGroup.Item as="li">
+        <LastUpdated label="Last updated at: " times={times} boldKey />
+      </ListGroup.Item>
       {allCurrentConditionsTimeseries.map((timeSeries, index) => {
         return <TableItem key={index} timeSeries={timeSeries} platform={platform} unitSystem={unitSystem} />
       })}
```
abkfenris, line 117: This feels like a weird overloading of `ERDDAPObervationTable` since there are two different sets of unrelated components depending on if a limit is set or not, which makes it a bit harder to follow.

Maybe refactor the `times` logic into a standalone function that can be shared and then have a similar `ERDDAPLimitedObservationTable`?

abkfenris (changes requested):

src/Features/ERDDAP/Platform/Observations/Table/table.tsx
shindelr: I agree, the coupling seems unnecessary and you're right, the times logic is really the only thing keeping what could be two separate components tied together.

shindelr:
Something that could affect the design choice here -- the spec says "Include top 2 conditions", that could be what we're doing here which is sort of random choice based on the slice with a limit, or I realized perhaps Dig actually wanted a `Superlatives` type of feature?

cgalvarino:
> Something that could affect the design choice here -- the spec says "Include top 2 conditions", that could be what we're doing here which is sort of random choice based on the slice with a limit, or I realized perhaps Dig actually wanted a `Superlatives` type of feature?

IIRC the obs arrive pre-sorted by priority that the UI doesn't know or care about.  So picking off the first 2 should == the top 2.

abkfenris:
> > Something that could affect the design choice here -- the spec says "Include top 2 conditions", that could be what we're doing here which is sort of random choice based on the slice with a limit, or I realized perhaps Dig actually wanted a `Superlatives` type of feature?
> 
> IIRC the obs arrive pre-sorted by priority that the UI doesn't know or care about. So picking off the first 2 should == the top 2.

They aren't sorted, just filtered to a subset of all possible timeseries to 'what is reasonably important to someone going out on the water' but not sorted within that subset.

shindelr:
> > > Something that could affect the design choice here -- the spec says "Include top 2 conditions", that could be what we're doing here which is sort of random choice based on the slice with a limit, or I realized perhaps Dig actually wanted a `Superlatives` type of feature?
> > 
> > 
> > IIRC the obs arrive pre-sorted by priority that the UI doesn't know or care about. So picking off the first 2 should == the top 2.
> 
> They aren't sorted, just filtered to a subset of all possible timeseries to 'what is reasonably important to someone going out on the water' but not sorted within that subset.

Hm, so in that case we may want to add some sorting functionality? By top conditions, are we just thinking biggest numbers?

cgalvarino:
I hope we can task the backend with that logic.

abkfenris:
> > > > Something that could affect the design choice here -- the spec says "Include top 2 conditions", that could be what we're doing here which is sort of random choice based on the slice with a limit, or I realized perhaps Dig actually wanted a `Superlatives` type of feature?
> > > 
> > > 
> > > IIRC the obs arrive pre-sorted by priority that the UI doesn't know or care about. So picking off the first 2 should == the top 2.
> > 
> > 
> > They aren't sorted, just filtered to a subset of all possible timeseries to 'what is reasonably important to someone going out on the water' but not sorted within that subset.
> 
> Hm, so in that case we may want to add some sorting functionality? By top conditions, are we just thinking biggest numbers?

I think that's probably a @tshyka @youngmorse question, but a straight numerical comparison would fall apart. A 10kt breeze is probably less important than a 9' swell, let alone a 16 kph breeze and a 2.74 meter swell.

We might want an adapted version of [`currentConditionsTimeseries`](https://github.com/gulfofmaine/Neracoos-1-Buoy-App/blob/8e4ac5e42a08740e3aa6b33797167ed7390255b0/src/Features/ERDDAP/utils/currentConditionsTimeseries.ts#L34-L73).

- Start with timeseries that are highlighted 'before'
- Order filtered conditions by most impactful (wave height, wind speed, wind gust, wave period...)

Then pick the top two.

cgalvarino:
Understood.  We'll add this to our punchlist for discussion.

cgalvarino:
#4078 instead

Issue #3879, opened 2026-01-08
MD Refresh - Station List/Regions


youngmorse:
**Station List Page - [https://mariners.neracoos.org/platform/I01](https://mariners.neracoos.org/platform/I01)**

The changes to this page include 
- Add new regions that have fewer stations (e.g. sub regions of GOM)
- Keep regions at same zoom level if possible
- Change bounding boxes for new regions

NERACOOS team to define new regions and bounding boxes.

**Station List Desktop**
<img width="969" height="1291" alt="Image" src="https://github.com/user-attachments/assets/c2304ff9-3a3f-46b8-9256-54e85bdeea09" />

**Quick Zoom to Next Region**

- Added links to go to the next closest region North and South of current region
- Loop the Northern-most region to to the Southern-most region and vice versa

**View Window**

- Cut off view window for station list at the bottom of the map and allow to scroll

**Link Color**

- Make link colors Blue

**Station Alert - may be Phase II**

- Add tag in station list for when a station has alert
  - feature may depend on updates to Buoy Barn to make warning/danger state available in API

**Mobile**

**Map Size**

- On mobile, make the collapsed view a square (height same as width) to ensure it doesn't take over the full screen

Station List Mobile
<img width="562" height="837" alt="Image" src="https://github.com/user-attachments/assets/26d1ad5e-c72c-44ba-acee-f85646eb299d" />

Station List Before
<img width="1054" height="1472" alt="Image" src="https://github.com/user-attachments/assets/a105de24-5942-4840-8983-1e87acd2d2b1" />

PR #4062, opened 2026-04-09
Scrollable region list & map height

shindelr:
@cgalvarino 

#3879 -- Scrollable station list
#3876 -- Map height minimum and match sidebar

Multiple important considerations in this PR.
Dig's specifications require different layout content height behavior on `/`, `/platform`, and `/region`.
- On mobile and all three routes: The map should collapse down to a small square.
- On `/`: Minimum map height should be 590px, made a specific CSS selector for the occasion to enable conditional styling. Otherwise, the map should grow to match sidebar content.
- On `/region`: Cap the height of the sidebar content so that a long list can scroll otherwise allow map to flex.
- On `/platform`: Map and sidebar content should both flex to grow/shrink together depending on the number of observations in the table. Conditional styling was needed here to prevent the scrolling and minimum height rules from applying. 

Specs - #3876
**Map Height**
Map height should match the height of the content on the right side
On Landing, map height minimum should be 590px

**Map Size**
On mobile, make the collapsed view a square (height same as width) to ensure it doesn’t take over the full screen

Specs - #3879 
**View Window**
Cut off view window for station list at the bottom of the map and allow to scroll

**Map Size**
On mobile, make the collapsed view a square (height same as width) to ensure it doesn't take over the full screen

**Before**
Screenshot 2026-04-09 at 6 09 44 PM

**After**
Screenshot 2026-04-09 at 6 08 53 PM

**Before - Mobile size**
Screenshot 2026-04-09 at 6 11 40 PM

**After - Mobile size**
Screenshot 2026-04-09 at 6 12 44 PM

### DIFF
--- a/app/(platformMap)/@sidebar/region/[regionId]/page.tsx
+++ b/app/(platformMap)/@sidebar/region/[regionId]/page.tsx
@@ -18,9 +18,10 @@ export default function RegionSidebar(props: { params: Promise<{ regionId: strin
   }
 
   return (
-    <div>
+    <div className="d-flex flex-column h-100">
       <h2>Platforms in {region.name}</h2>
-      <div className="row mb-2 align-items-center">
+
+      <div className="d-flex flex-row mb-2 align-items-center">
         <div className="d-flex col-6">
           <NextRegion region={region} offset={-1} />
         </div>
@@ -28,9 +29,12 @@ export default function RegionSidebar(props: { params: Promise<{ regionId: strin
           <NextRegion region={region} offset={1} />
         </div>
       </div>
-      <DehydratedPlatforms>
-        <RegionList region={region} />
-      </DehydratedPlatforms>
+
+      <div className="overflow-y-auto border border-1 rounded">
+        <DehydratedPlatforms>
+          <RegionList region={region} />
+        </DehydratedPlatforms>
+      </div>
     </div>
   )
 }
--- a/app/(platformMap)/@sidebar/region/[regionId]/page.tsx
+++ b/app/(platformMap)/@sidebar/region/[regionId]/page.tsx
@@ -18,7 +18,7 @@ export default function RegionSidebar(props: { params: Promise<{ regionId: strin
   }
 
   return (
-    <div className="d-flex flex-column h-100">
+    <div className="d-flex flex-column region-list-container">
       <h2>Platforms in {region.name}</h2>
 
       <div className="d-flex flex-row mb-2 align-items-center">
@@ -30,7 +30,7 @@ export default function RegionSidebar(props: { params: Promise<{ regionId: strin
         </div>
       </div>
 
-      <div className="overflow-y-auto border border-1 rounded">
+      <div className="region-list border border-1 rounded">
         <DehydratedPlatforms>
           <RegionList region={region} />
         </DehydratedPlatforms>
--- a/app/(platformMap)/layout.tsx
+++ b/app/(platformMap)/layout.tsx
@@ -2,7 +2,6 @@
 import { useParams, usePathname } from "next/navigation"
 import * as React from "react"
 import { useEffect, useState } from "react"
-import { useMeasure } from "react-use"
 import Col from "react-bootstrap/Col"
 import Row from "react-bootstrap/Row"
 
@@ -22,7 +21,6 @@ export default function Layout({
   const path = usePathname()
   const isPlatformView = path.startsWith("/platform")
   const params: { regionId?: string; platformId?: string } = useParams()
-  let [ref, { height }] = useMeasure<HTMLDivElement>()
   const platformId = params.platformId
   const [isClient, setIsClient] = useState(false)
 
--- a/app/(platformMap)/layout.tsx
+++ b/app/(platformMap)/layout.tsx
@@ -20,6 +20,7 @@ export default function Layout({
 }) {
   const path = usePathname()
   const isPlatformView = path.startsWith("/platform")
+  const isRegionView = path.startsWith("/region")
   const params: { regionId?: string; platformId?: string } = useParams()
   const platformId = params.platformId
   const [isClient, setIsClient] = useState(false)
@@ -37,14 +38,19 @@ export default function Layout({
 
   return (
     <React.Fragment>
-      <Row className="g-5">
-        <Col xs={12} md={6} className="order-2">
-          <div className="platform-map-layout">{sidebar}</div>
+      <Row className="g-5 align-items-stretch">
+        <Col xs={12} md={6} className="order-2 d-flex">
+          <div className="platform-map-layout flex-fill">
+            <div className={isRegionView ? "h-60" : ""}>{sidebar}</div>
+          </div>
         </Col>
 
-        <Col xs={12} md={6} className="order-1">
-          <div className="platform-map-layout">
-            <ErddapMap className="map" {...(isPlatformView && { platformId })} />
+        <Col xs={12} md={6} className="order-1 d-flex flex-column">
+          <div className="platform-map-layout flex-fill">
+            <ErddapMap
+              className={`${!(isPlatformView || isRegionView) ? "min-590px" : ""} map`}
+              {...(isPlatformView && { platformId })}
+            />
           </div>
 
           {/* Below Map = Superlatives */}
--- a/app/(platformMap)/layout.tsx
+++ b/app/(platformMap)/layout.tsx
@@ -42,7 +42,7 @@ export default function Layout({
         <Col xs={12} md={6} className="order-2 d-flex">
           <div className="platform-map-layout flex-fill">
             {/* Second div layer decouples sidebar from map to allow for scrolling */}
-            <div className={isRegionView ? "h-60" : ""}>{sidebar}</div>
+            <div className={isRegionView ? "sidebar-height" : ""}>{sidebar}</div>
           </div>
         </Col>
 
--- a/app/(platformMap)/layout.tsx
+++ b/app/(platformMap)/layout.tsx
@@ -39,9 +39,7 @@ export default function Layout({
     <React.Fragment>
       <Row className="g-5">
         <Col xs={12} md={6} className="order-2">
-          <div className="platform-map-layout">
-            {sidebar}
-          </div>
+          <div className="platform-map-layout">{sidebar}</div>
         </Col>
 
         <Col xs={12} md={6} className="order-1">
--- a/app/(platformMap)/layout.tsx
+++ b/app/(platformMap)/layout.tsx
@@ -30,10 +30,6 @@ export default function Layout({
     setIsClient(true)
   }, [])
 
-  if (height < 420) {
-    height = 420
-  }
-
   //Removing warning in console re:Highcharts library defaultProps. Not an active warning for us--until this issues is solved on Highcharts' end, keep this to remove huge console error
   const error = console.error
   console.error = (...args: any) => {
@@ -45,11 +41,15 @@ export default function Layout({
     <React.Fragment>
       <Row className="g-5">
         <Col xs={12} md={6} className="order-2">
-          <div ref={ref}>{sidebar}</div>
+          <div className="platform-map-layout" ref={ref}>
+            {sidebar}
+          </div>
         </Col>
 
         <Col xs={12} md={6} className="order-1">
-          <ErddapMap height={params.regionId ? "80vh" : height} {...(isPlatformView && { platformId })} />
+          <div className="platform-map-layout">
+            <ErddapMap className="map" {...(isPlatformView && { platformId })} />
+          </div>
           {belowMap ?? <React.Fragment>{belowMap}</React.Fragment>}
         </Col>
       </Row>
--- a/app/(platformMap)/layout.tsx
+++ b/app/(platformMap)/layout.tsx
@@ -46,11 +46,16 @@ export default function Layout({
           <div className="platform-map-layout">
             <ErddapMap className="map" {...(isPlatformView && { platformId })} />
           </div>
+
+          {/* Below Map = Superlatives */}
           {belowMap ?? <React.Fragment>{belowMap}</React.Fragment>}
         </Col>
       </Row>
 
-      {(isPlatformView && isClient && bottom) ?? <Row>{bottom}</Row>}
+      {/* Bottom = Platform obs data */}
+      <Row className="mt-4 d-flex justify-content-center">
+        {(isPlatformView && isClient && bottom) ?? <React.Fragment>{bottom}</React.Fragment>}
+      </Row>
     </React.Fragment>
   )
 }
--- a/app/(platformMap)/layout.tsx
+++ b/app/(platformMap)/layout.tsx
@@ -41,6 +41,7 @@ export default function Layout({
       <Row className="g-5 align-items-stretch">
         <Col xs={12} md={6} className="order-2 d-flex">
           <div className="platform-map-layout flex-fill">
+            {/* Second div layer decouples sidebar from map to allow for scrolling */}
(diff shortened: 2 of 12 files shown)


cgalvarino:
Great work.  I still have bad feels about the footer, thought.

This extra space and it still scrolls?
image
<br/>
If I look at the spec, I wonder if this is a sticky footer?
image

cgalvarino:
And thx for adding the spec info to the PR.  Makes my job so much easier!

shindelr:
I see what you're saying there and I believe you are correct. Current white space is an artifact of old styling, I think 110px of padding/margin on the bottom of the body. I'll tinker with removing that and stickifying the footer

cgalvarino:
Dern.  Looks like an issue for horizontal mobile.  Map clipping.

Prod
image

This PR
image

shindelr:
Hm good catch, is the goal with horizontal mobile to fit the whole map and sidebar on the screen without scrolling? It's looking like the landscape mode is triggering the desktop/tablet sizing rules which limits minimum map height to 590px -- doesn't fit on a 350px screen

shindelr:
No min height and it's great on landscape mobile but then no hard rule as given in the specs. In my mind, perhaps it'd be better have things able to flex rather than get caught up in static height rules?

Screenshot 2026-04-15 at 11 05 23 AM

shindelr:
@cgalvarino give this a run through. Should solve our landscape mobile problem as well! Had to add some conditional rendering to allow the single layout to shift and flex appropriately given the different requirements on /, /platform, and /region

cgalvarino:
Wow, looks great.  Your conditional rendering leaves Water Level alone, too.  And allows us to save the station pages for a coming attraction.

Would you mind updating the original PR comment w/ any notes that a future developer might want to know, e.g., why you care about /regions and what that means for the height(s)?  @abkfenris may want to see some of this as in-line comments.

Other than that, @abkfenris plz take a peek for merge.

src/Features/ERDDAP/Map/index.tsx
```diff
@@ -169,16 +171,10 @@ export const ErddapMapBase: React.FC<BaseProps> = ({ platforms, platformId, heig
     }
   }
 
-  // Make sure the height of the map gets updated when jumping
-  // from home to platform view
-  useLayoutEffect(() => {
-    mapRef?.current?.ol.updateSize()
```
cgalvarino, line 175: This is what I was talking about . . . the openlayers update-yourself callback.  You sure we don't need it?

shindelr: I think we're good to go since we're explicitly defining height at 100% at all times in index.scss and then letting the row and parent containers control the flex. The kicker that was breaking my brain was figuring out how to make the sidebar and map containers flex on station details while still defining the siderbar height so it could scroll conditionally and disallowing the map from shrinking conditionally!

app/(platformMap)/@sidebar/region/[regionId]/page.tsx
```diff
@@ -18,19 +18,23 @@ export default function RegionSidebar(props: { params: Promise<{ regionId: strin
...
       </div>
-      <DehydratedPlatforms>
-        <RegionList region={region} />
-      </DehydratedPlatforms>
+
+      <div className="overflow-y-auto border border-1 rounded">
+        <DehydratedPlatforms>
+          <RegionList region={region} />
+        </DehydratedPlatforms>
+      </div>
```
Comment (abkfenris) [inline app/(platformMap)/@sidebar/region/[regionId]/page.tsx:37]: I think this causes a double scroll on mobile which is a weird feeling interaction, especially since it doesn't necessarily mean that the map is kept fully on screen anyways. 

Image

There are also not necessarily great indicators that the list is scroll-able.

The sticky footer is also a waste of usable screen real estate on mobile.

Comment (abkfenris) [reply app/(platformMap)/@sidebar/region/[regionId]/page.tsx:37]: I also just tried horizontal mobile and the map is nowhere to be found.

Image

Comment (abkfenris) [reply app/(platformMap)/@sidebar/region/[regionId]/page.tsx:37]: The map is also missing from platform pages at iPhone SE horizontal dims, and iPad mini vertical dims.

Image

Image

app/(platformMap)/layout.tsx
```diff
@@ -43,18 +38,32 @@ export default function Layout({
...
+            <ErddapMap
+              // Pass minimum height requirement only on landing page
+              className={`${!(isPlatformView || isRegionView) ? "landing-min-height" : ""} map`}
+              {...(isPlatformView && { platformId })}
+            />
+          </div>
+
+          {/* Below Map = Superlatives */}
           {belowMap ?? <React.Fragment>{belowMap}</React.Fragment>}
         </Col>
```
abkfenris, line 60: Comments with images are threaded elsewhere, but at some phone horizontal and iPad mini vertical dims, the map is missing.

abkfenris (changes requested):
I also just noticed 'Station List' in place of 'Regions'. I feel like that should just be 'Stations' and then have an alt text along the lines of 'station list for each region'.

app/(platformMap)/@sidebar/region/[regionId]/page.tsx
Comment (cgalvarino) [reply app/(platformMap)/@sidebar/region/[regionId]/page.tsx:37]: > I think this causes a double scroll on mobile which is a weird feeling interaction, especially since it doesn't necessarily mean that the map is kept fully on screen anyways.
> 
> Image
> There are also not necessarily great indicators that the list is scroll-able.
> 
> The sticky footer is also a waste of usable screen real estate on mobile.

I agree that the list should not scroll on mobile, but the spec indicates otherwise.  From https://github.com/gulfofmaine/Neracoos-1-Buoy-App/issues/3879.

image

shindelr:
> I also just noticed 'Station List' in place of 'Regions'. I feel like that should just be 'Stations' and then have an alt text along the lines of 'station list for each region'.

Happy to go with that, though specs asked for 'Station List', if you check out #3876

app/(platformMap)/@sidebar/region/[regionId]/page.tsx
Comment (abkfenris) [reply app/(platformMap)/@sidebar/region/[regionId]/page.tsx:37]: Unless @youngmorse or @tshyka say otherwise, I vote we ignore the specs on that. I especially don't trust specs that don't match any real device dimensions.

Comment (shindelr) [reply app/(platformMap)/@sidebar/region/[regionId]/page.tsx:37]: Scrollable list on mobile was little odd, I agree. Though super long station list on mobile is a little funny too. I did notice in that wireframe just now that the prev/next buttons are at the top of the map and current implementation has them between map and list.

app/(platformMap)/layout.tsx
shindelr: Interesting! I've been testing mobile horizontal on iPhone XR and Galaxy Z Fold 5. Looks like iPhone SE has both a small width and height on horizontal which is causing map and sidebar to stack without rendering map. Looking into a solution now

shindelr:
@abkfenris @cgalvarino If you could check out the change I made. Adjusted the breakpoint so that the map doesn't go without a min-height. iPhone SE has a horizontal map, but since the phone's length is just under the medium breakpoint we end up with a mobile (stacked) layout but with a mega map maxed out to the width (which is actually the length) of the phone.

shindelr:
Added a limit to the mobile max height at 50vh so that you can get your fingers in there and at least see that there's more content below. Shouldn't affect portrait mode since the map's size is constrained to 1/1 ratio controlled by the width

Screenshot 2026-04-20 at 12 28 31 PM

shindelr:
@cgalvarino Adjusted sidebar height to take up 80vh of the screen, removed the padding/margin between footer and main body content, and removed scrollability on mobile portrait mode

cgalvarino:
When I looked at this last, I thought the map was filling up all empty space.  Doesn't always seem to be the case now.

image

shindelr:
@cgalvarino finally, a panacea to end all woe. Added a media query for screen height sub 500px, to let the body content shrink more rapidly. Allow sidebar to flex up to a max of 85% of the screen for those taller than 500px.

Screenshot 2026-04-20 at 7 53 34 PM

Screenshot 2026-04-20 at 7 54 00 PM

cgalvarino:
This looks great.  Just to make sure that I know what this PR should and should not address . . . the regions list pages look great for all devices and orientations.  Is it true that the landing page does NOT get this love and that it will be addressed when you move the top obs content over?

iphone 12 pro
image

shindelr:
Landing page map has a min-height of 590px per the specs, so definitely larger than a landscape phone.

Moving the OBS content to the right probably won't solve that problem unless we go off spec and let things shrink

cgalvarino:
OK.  Thanks.  Just wanted verification.  If the group decides that needs tweaking, we'll tweak it then.

We'll == you'll

@abkfenris this one's all yours.

app/(platformMap)/layout.tsx
abkfenris: It looks like the map is showing on all dims/layouts I've tried.

app/(platformMap)/@sidebar/region/[regionId]/page.tsx
Comment (abkfenris) [reply app/(platformMap)/@sidebar/region/[regionId]/page.tsx:37]: I'd rather see a long station list when stacked for small widths than to have a scrollable list with no indications that scrolling is necessary. A bunch of standard device sizes give no (or 1px) hints that it's scrollable.

Image

Image

Image

Comment (shindelr) [reply app/(platformMap)/@sidebar/region/[regionId]/page.tsx:37]: Looks like we lost portrait mode _un_scrollability on the region list. Troubleshooting that now

Comment (shindelr) [reply app/(platformMap)/@sidebar/region/[regionId]/page.tsx:37]: @abkfenris check it out, we now have no scroll on mobile portrait, but scroll on medium breakpoints and up. 

Map is shrinking/growing to match sidebar content on the region list. Not sure if you and @cgalvarino like that, or would like the map to stay a constant size. Let me know

Comment (abkfenris) [reply app/(platformMap)/@sidebar/region/[regionId]/page.tsx:37]: That feels much better, thank you.

cgalvarino:
This all seems like a good UX to me.

This has turned into quite the beast.  How about we freeze dev, and you or I demo it to the group tomorrow.  If we get buy-in, fingers crossed, we can then merge it an move on!

abkfenris (approved):
This looks good. 

If you land before the UMaine issues are fixed, then the dev deployment will fail and it won't be visible before we have fresh data.

---

One thing that I noticed since we've been poking around the region pages so much in this PR (and there was a vaguely related comment), is the ordering of the regions for the previous and next seems backwards. 

The `regionList` that defines the menu, goes roughly North > South, which corresponds East > West. 

https://github.com/shindelr/Neracoos-1-Buoy-App/blob/9db4397cdf7764f34303995c560d378801ac62c0/src/Shared/regions.ts#L117-L126

I got codespaces into a weird state trying to make this look more like what my brain expects, but here's the diff of my tinkering.

```diff
commit 1ae07bd3dfe0cef83041085847946255b5ba0ab0 (HEAD -> feature-scrollable-region-list)
Author: Alex Kerney <abk@mac.com>
Date:   Tue Apr 21 18:41:22 2026 +0000

    Reverse region arrow links.

diff --git a/app/(platformMap)/@sidebar/region/[regionId]/page.tsx b/app/(platformMap)/@sidebar/region/[regionId]/page.tsx
index a33748f8..b8747fdb 100644
--- a/app/(platformMap)/@sidebar/region/[regionId]/page.tsx
+++ b/app/(platformMap)/@sidebar/region/[regionId]/page.tsx
@@ -23,10 +23,10 @@ export default function RegionSidebar(props: { params: Promise<{ regionId: strin
 
       <div className="d-flex flex-row mb-2 align-items-center">
         <div className="d-flex col-6">
-          <NextRegion region={region} offset={-1} />
+          <NextRegion region={region} offset={1} />
         </div>
         <div className="d-flex col-6 justify-content-end">
-          <NextRegion region={region} offset={1} />
+          <NextRegion region={region} offset={-1} />
         </div>
       </div>
 
diff --git a/app/(platformMap)/@sidebar/region/[regionId]/region.tsx b/app/(platformMap)/@sidebar/region/[regionId]/region.tsx
index e2c967cc..b158d1cb 100644
--- a/app/(platformMap)/@sidebar/region/[regionId]/region.tsx
+++ b/app/(platformMap)/@sidebar/region/[regionId]/region.tsx
@@ -31,9 +31,9 @@ export function NextRegion({ region, offset }: { region: Region; offset: number
         key={regionList[nextRegionIdx].name}
         className="d-flex flex-row align-items-center gap-1"
       >
-        {offset < 0 && <ArrowLeftIcon className="text-primary fa-md" />}
+        {offset > 0 && <ArrowLeftIcon className="text-primary fa-md" />}
         <span className="text-center">{regionList[nextRegionIdx].name}</span>
-        {offset > 0 && <ArrowRightIcon className="text-primary fa-md" />}
+        {offset < 0 && <ArrowRightIcon className="text-primary fa-md" />}
       </Link>
     </>
   )
```

Image

vs

Image

cgalvarino:
> This looks good.
> 
> If you land before the UMaine issues are fixed, then the dev deployment will fail and it won't be visible before we have fresh data.
> 
> One thing that I noticed since we've been poking around the region pages so much in this PR (and there was a vaguely related comment), is the ordering of the regions for the previous and next seems backwards.
> 
> The `regionList` that defines the menu, goes roughly North > South, which corresponds East > West.
> 
> https://github.com/shindelr/Neracoos-1-Buoy-App/blob/9db4397cdf7764f34303995c560d378801ac62c0/src/Shared/regions.ts#L117-L126
> 
> I got codespaces into a weird state trying to make this look more like what my brain expects, but here's the diff of my tinkering.
> 
> ```diff
> commit 1ae07bd3dfe0cef83041085847946255b5ba0ab0 (HEAD -> feature-scrollable-region-list)
> Author: Alex Kerney <abk@mac.com>
> Date:   Tue Apr 21 18:41:22 2026 +0000
> 
>     Reverse region arrow links.

Haha.  I didn't even think about that.  Pulldown puts you N-S.  Arrow nav puts you in the W-E mindset.

What if your suggested fix doesn't hold true once all the new regions have been defined?  Maybe we could have a second regionList . . . regionListPrevNext . . . that is assigned to the horizontal nav.  That way there'd be no guessing?

abkfenris:
> > This looks good.
> > If you land before the UMaine issues are fixed, then the dev deployment will fail and it won't be visible before we have fresh data.
> > One thing that I noticed since we've been poking around the region pages so much in this PR (and there was a vaguely related comment), is the ordering of the regions for the previous and next seems backwards.
> > The `regionList` that defines the menu, goes roughly North > South, which corresponds East > West.
> > https://github.com/shindelr/Neracoos-1-Buoy-App/blob/9db4397cdf7764f34303995c560d378801ac62c0/src/Shared/regions.ts#L117-L126
> > I got codespaces into a weird state trying to make this look more like what my brain expects, but here's the diff of my tinkering.
> > ```diff
> > commit 1ae07bd3dfe0cef83041085847946255b5ba0ab0 (HEAD -> feature-scrollable-region-list)
> > Author: Alex Kerney <abk@mac.com>
> > Date:   Tue Apr 21 18:41:22 2026 +0000
> > 
> >     Reverse region arrow links.
> > ```
> 
> Haha. I didn't even think about that. Pulldown puts you N-S. Arrow nav puts you in the W-E mindset.
> 
> What if your suggested fix doesn't hold true once all the new regions have been defined? Maybe we could have a second regionList . . . regionListPrevNext . . . that is assigned to the horizontal nav. That way there'd be no guessing?

A separate array would work, we might just want to toss a test in that makes sure that both arrays include all regions, because I know what happens as soon as we start adding more/changing order.

cgalvarino:
@shindelr gets all the glory.  I'll take all the blame.

This trumps my idea of waiting til tomorrow:
image

PR #4068, opened 2026-04-13
Temperature unit abbreviation

shindelr:
@cgalvarino
#3875 -- Abbreviate units everywhere

Added abbreviated units to temperature class. 
Hex code \u00B0 == °

**Before**
Screenshot 2026-04-13 at 1 45 55 PM

**After**
Screenshot 2026-04-13 at 1 46 37 PM


tsconfig.json
```diff
@@ -9,7 +9,7 @@
     "strict": true,
     "forceConsistentCasingInFileNames": true,
     "module": "esnext",
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
```
cgalvarino, line 12: Not sure what this is.  Danger zone?

cgalvarino:
If my danger zone is a false flag, this looks good to me, @abkfenris .

tsconfig.json
shindelr: I've been having trouble with this! Yesterday when I ran `make up`, I got a nextjs message in the terminal saying that this change was "mandatory". Manually discarded the change since my danger zone flag went off too but it keeps changing?

abkfenris: Probably some change that didn't get included with the last NextJS security update.

abkfenris (approved):

PR #4069, opened 2026-04-14
Cardinal direction abbreviation

shindelr:
@cgalvarino
https://github.com/gulfofmaine/Neracoos-1-Buoy-App/issues/3875 -- Abbreviate units everywhere

Changed "Degrees" to "°"
Hex code \u00B0 == °

**Before**
Screenshot 2026-04-14 at 12 32 49 PM

**After**
Screenshot 2026-04-14 at 12 32 18 PM

### DIFF
--- a/next-env.d.ts
+++ b/next-env.d.ts
@@ -1,6 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
-import "./.next/dev/types/routes.d.ts"
+import "./.next/dev/types/routes.d.ts";
 
 // NOTE: This file should not be edited
 // see https://nextjs.org/docs/app/api-reference/config/typescript for more information.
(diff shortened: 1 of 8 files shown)


cgalvarino:
I'm not sure where to see this.

Is it supposed to trickle down to the chart mouseovers?
image

shindelr:
Only shows up when the unit system is in metric. Screenshots above are from the table:
Screenshot 2026-04-14 at 1 48 30 PM

Interestingly, the chart mouseover seems to only use "English" style directions
Screenshot 2026-04-14 at 1 49 16 PM

But if you go to the specific chart it respects the different unit systems
Screenshot 2026-04-14 at 1 50 44 PM

shindelr:
Also, how do you feel about the whitespace between the degree symbol and the number?

cgalvarino:
Ok. Odd. But it doesn’t matter what I think. @abkfenris reqdy for merge if you are.

abkfenris (approved):
The whitespace is a little weird, but I'm not sure if there is a way to cleanly do that without adding a method like `formatValue()` that can understand if whitespace is necessary or not.

PR #4080, opened 2026-04-18
one popup per map instead of one popup per feature

cgalvarino:
Enough clobbering tooltips, already!

Before: One RPopup per feature.

After:  One RPopup per map.  The map now has an onPointerMove listener.  This listener is smart enough to pick the topmost platform dot to use for the RPopup guts.

@shindelr @abkfenris

### DIFF
--- a/src/Features/ERDDAP/Map/index.tsx
+++ b/src/Features/ERDDAP/Map/index.tsx
@@ -105,27 +105,11 @@ export const PlatformLayer = ({ platform, selected, old = false }: PlatformLayer
       </RStyle.RStyle>
       <RFeature
         geometry={geometry}
+        properties={{ platform: platform, url: url }}
         onClick={useCallback(() => {
           router.push(url)
         }, [router, url])}
-      >
-        <RPopup trigger={"hover"}>
-          <Button
-            variant="dark"
-            size="sm"
-            href={url}
-            onClick={useCallback(
-              (event) => {
-                event.preventDefault()
-                router.push(url)
-              },
-              [router, url],
-            )}
-          >
-            {platformName(platform)}
-          </Button>
-        </RPopup>
-      </RFeature>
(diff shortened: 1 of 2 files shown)


src/Features/ERDDAP/Map/index.tsx
```diff
@@ -105,27 +105,11 @@ export const PlatformLayer = ({ platform, selected, old = false }: PlatformLayer
...
-            size="sm"
-            href={url}
-            onClick={useCallback(
-              (event) => {
-                event.preventDefault()
-                router.push(url)
-              },
-              [router, url],
-            )}
-          >
```
cgalvarino, line 127: @abkfenris I didn't bring over the `onClick` callback to the new RPopup.Button.  It seems the that `href` that I did bring over does the trick.  Am I missing out on some functionality w/o the `onClick`?

The feature still has the `onClick`.

abkfenris: I think that was due to a difference in targets. If I'm remembering correctly, the `href` made the contents a link, but if you hit the background it needed the `onClick` instead.

cgalvarino: That makes sense.  In this implementation the tooltip is associated w/ a temporary feature that sits under the main feature.  So the main feature is the one receiving the events.

cgalvarino:
@abkfenris @shindelr This is ready for prime time AFAICT.

I noticed some edge cases.  Quite literally.  But this same behavior is on prod in clobberville.  I.e., I didn't introduce it here, so it ain't a new bug.

If you click juuuuuuust to the top left of the tooltip, nothing fires.  I'm in favor of calling this PR done and saving that fix for down the road when we have time to kill.

image

abkfenris (approved):

PR #4084, opened 2026-04-20
Top OBS info icon & map icon size

shindelr:
@cgalvarino 
#3876 

Adjusted map icon size to be 44px x 44px on mobile.
Added info icon to top wind and waves card on landing which displays last updated time stamp.

**Icon Size**
On mobile, increase size of icons on map to be 44x44 px to ensure large enough touch target

**Latest Conditions**
Renamed title to ‘Top Wind & Waves—All Regions’ (H3)
Moved to the right of the map
**Added info icon popup to house Last Updated timestamp and other information**
Background color: Black 5
Text styles (in order): ’Paragraph Bold’, ‘Paragraph’, ‘Link’
Title & Paragraph color: Black
Link color: Blue

**Before**
Screenshot 2026-04-20 at 4 01 21 PM

**After**
Screenshot 2026-04-20 at 4 00 53 PM

### DIFF
--- a/src/Features/ERDDAP/Superlatives/index.tsx
+++ b/src/Features/ERDDAP/Superlatives/index.tsx
@@ -8,6 +8,7 @@ import { useEffect, useState } from "react"
 import Card from "react-bootstrap/Card"
 import Col from "react-bootstrap/Col"
 import Row from "react-bootstrap/Row"
+import { Tooltip, OverlayTrigger } from "react-bootstrap"
 
 import { WarningAlert } from "components/Alerts"
 import { useUnitSystem } from "Features/Units"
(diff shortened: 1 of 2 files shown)


cgalvarino:
I guess the contents are according to spec, but the tooltip text seems really odd.  But I think the order should match the obs:  winds then waves.

shindelr:
Changed the order of the wind and waves observations. 

The specs also say "other information" so maybe we can spruce this hover up in a later PR?

cgalvarino:
Reorder looks good.

The icon size on mobile, though.  Since you're relying on breakpoints, we don't see the gollywhopper icons on landscape mobile.

Vertical OK.
image

Landscape not 44x44.
image

If you take out the `@include media-breakpoint-down(md) {` wrapper, it seems to work fine.  Desktop doesn't use this class.  Aside;  I wonder if there's a secret there to help out w/ the whole scrollable region list.

shindelr:
Whoops, guess I was overthinking that one!

Resolved now though.
Screenshot 2026-04-21 at 9 15 00 PM

cgalvarino:
Jeez, those are some big a$$ buttons.

abkfenris (approved):
> Jeez, those are some big a$$ buttons.

At least we're only going for thumb sized buttons. There might not be a map left otherwise.

PR #4091, opened 2026-04-24
Region list map height fix

shindelr:
@cgalvarino 
#3879

Map height flexes with viewport but stays constant on region list.

Screenshot 2026-04-24 at 9 55 34 AM

Screenshot 2026-04-24 at 9 55 26 AM


cgalvarino:
Looks great.  @abkfenris

abkfenris (approved):

PR #4095, opened 2026-04-24
Map drag bugfix for clobberville

shindelr:
@cgalvarino 

Bug fix for the map drag re-render problem for the non-clobbering popups.

Notable changes from OG implementation:
- **highlightedFeatures: any[] -> highlightedFeature: any**: Figured if we're only putting one value in there, probably doesn't need to be an array.
- **Added highlightedFeatureRef**: Need this to keep track of and what feature is hovered over and change the state on highlightedFeature without doing it constantly.
- **Moved the popup show and hide logic to useEffect**: Forces the map to re-render and show/hide the popup whenever highlightedFeature changes, but only when it changes.

### DIFF
--- a/src/Features/ERDDAP/Map/index.tsx
+++ b/src/Features/ERDDAP/Map/index.tsx
@@ -27,6 +27,7 @@ import { PlatformFeature } from "../types"
 import { platformName } from "../utils/platformName"
 import { Feature } from "ol"
 import { StationPopup } from "./StationPopup"
+import { feature } from "@turf/helpers"
 
 export interface Props {
   // Bounding box for fitting to a region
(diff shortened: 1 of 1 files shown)


cgalvarino:
Man, I'm willing this to work so bad, but I'm running into some weird panning effects.

https://app.screencast.com/KwH8SOwg4i0lj?conversation=LV04evcjasz1oNZTtMnWvF

Since this no-clobber was not on the spec, I don't think we can afford to spend anymore time on it.  At least not this month.  What say you both that I roll back my original no-clobber and revisit this later when / if we have time?

shindelr:
Darn! I agree, let's table this until we have some down time

PR #4096, opened 2026-04-24
Map move

cgalvarino:
@shindelr wanna try this?

### DIFF
--- a/src/Features/ERDDAP/Map/index.tsx
+++ b/src/Features/ERDDAP/Map/index.tsx
@@ -27,6 +27,7 @@ import { PlatformFeature } from "../types"
 import { platformName } from "../utils/platformName"
 import { Feature } from "ol"
 import { StationPopup } from "./StationPopup"
+import { feature } from "@turf/helpers"
 
 export interface Props {
   // Bounding box for fitting to a region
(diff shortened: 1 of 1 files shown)


shindelr:
Yes this great! I just moved the map around a ton without any problems. So many refs and hooks for what I would've expected to be a simple problem

shindelr (approved):
Looks good to me! At some point we may want to come through and add types to everything

cgalvarino:
Teamwork makes the dreamwork.  @abkfenris

abkfenris:
Uhh, I'm unsure what problem you're trying to solve here, so not sure what I should be looking for.

cgalvarino:
If you go to dev, panning the map is completely busted.  Seems like the no-clobber popups are causing it to rerender unwelcomingly.  @shindelr's description:

Notable changes from OG implementation:
- **highlightedFeatures: any[] -> highlightedFeature: any**: Figured if we're only putting one value in there, probably doesn't need to be an array.
- **Added highlightedFeatureRef**: Need this to keep track of and what feature is hovered over and change the state on highlightedFeature without doing it constantly.
- **Moved the popup show and hide logic to useEffect**: Forces the map to re-render and show/hide the popup whenever highlightedFeature changes, but only when it changes. 

And I added a couple of listeners to prevent any tooltip action if the map is moving.

src/Features/ERDDAP/Map/index.tsx
```diff
@@ -161,18 +183,15 @@ export const ErddapMapBase: React.FC<BaseProps> = ({ platforms, platformId, clas
 
     // Iterate through all features at the current pixel
     map.forEachFeatureAtPixel(e.pixel, (feature) => {
```
abkfenris, line 185: Is this shadowing the imported feature, or is that imported feature unused?

abkfenris (approved):
Thanks for the explanation.

And it's not the first time the map has decided that it's favorite thing is to constantly re-render.

src/Features/ERDDAP/Map/index.tsx
cgalvarino: I'm not quite sure what you're asking.

Every mousemove will pluck top-most feature at the cursor and essentially clone it so that it lives in this 0-pixel vector here, https://github.com/cgalvarino/Neracoos-1-Buoy-App/blob/32b5d2047ece94c706c0da057fd768cbdc963fde/src/Features/ERDDAP/Map/index.tsx#L248.  And all that bad boy does is show or hide its attached popup.

Issue #3877, opened 2026-01-08
MD Refresh - Station Detail Page


youngmorse:
(issue in progress)

Station Detail Page - [https://mariners.neracoos.org/platform/I01](https://mariners.neracoos.org/platform/I01)

- [ ] Map
- [ ] Station Attribution
- [ ] Latest Data
- [ ] Tab Bar
- [ ] 7 Day Plots

PR #4098, opened 2026-05-04
Station detail latest conditions cards

shindelr:
@cgalvarino 
#3877 -- Latest conditions cards

Migrates from list oriented latest conditions to cards. Turned out to be a fairly significant construction to get things to spec. The majority of the complexity comes from needing two special cards for Winds and Waves followed by any number of "generic" cards for other data such as air temperature.

The Wind and Wave cards combine multiple timeseries into a single card with a bit of conditional formatting to add in language like "Gust:" or "Period:".  98% of the logic for this is handled in `latestObs.ts` which I tried to make as abstract as I could since I figured we could try to group other bits of data to make more multi-data cards.

On the radar not implemented in this initial PR:
- Passthrough units abbreviated, like Seconds.

**Goal**
Screenshot 2026-05-04 at 6 27 33 PM

**After**
Screenshot 2026-05-04 at 6 30 29 PM

**Before**
Screenshot 2026-05-04 at 6 31 15 PM


src/Features/ERDDAP/Platform/Observations/CurrentConditions/index.tsx
abkfenris: I think we should adapt the `currentConditionsTimeseries` function for the latest conditions card, so that we have one place where we're managing highlighted timeseries.

abkfenris (commented):
Not fully digging into things yet, just snooping around and immediately got a bit of a weird state when trying things.

Image

It doesn't feel right to me to loose the conditions summary when digging into a specific data type.

Image

@tshyka @youngmorse  It feels to me like we should promote that back up below station info, because right now it's just showing the current conditions mini charts another way, but only on that page.

shindelr:
Good notes, Alex. Still have some loose ends to tie up like the button group there.

Just my two cents on the location of the latest conditions, I think we might be looking at two different user stories here. One user might want to grab a high level number like wind speed or wave height in as little time as possible. Another user might want to dig into data and consider trends/patterns. Are users who are looking for quick facts also looking at the big charts in the same session?

src/Features/ERDDAP/Platform/Observations/CurrentConditions/index.tsx
shindelr: This makes sense to me, so you're thinking move some of the logic out of latestObs.ts to here?

cgalvarino:
Gents, let me get a chance to talk w/ Tom and Riley about the specs.  I want to avoid any wheel spinning if possible.

Having said that . . . what if clicking on a mini graph caused the area in red to maximize to the entire screen?  Then it's pretty clear that the original screen w/ the latest obs is still available underneath it.

image

abkfenris:
I think it's a balance of progressive disclosure and balancing context.

- Some users will get enough info from hovering around the map now that some observations will be shown.
- Many users will want to see a trend of multiple data types so they end up at the current conditions page.
- Some users will want to see further, so they will go to a single data type. While some of those may just be interested in the past, many will still be thinking about how it affects the current conditions so they want the context of what else is going on in the latest conditions.

I'd toss the filtering buttons. They add more UX noise than they potentially clean up with the limited types of data being displayed.

> Gents, let me get a chance to talk w/ Tom and Riley about the specs. I want to avoid any wheel spinning if possible.
> 
> Having said that . . . what if clicking on a mini graph caused the area in red to maximize to the entire screen? Then it's pretty clear that the original screen w/ the latest obs is still available underneath it.
> image

How would this act for data types that aren't on the current conditions page?

src/components/PlatformInfo/platformInfo.tsx
```diff
@@ -3,31 +3,18 @@ import React from "react"
...
-  const unitSystem = useUnitSystem()
-  const aDayAgo = aDayAgoRounded()
-
   return (
     <UsePlatform platformId={id}>
       {({ platform }) => (
         <React.Fragment>
           <PlatformAlerts platform={platform} />
           <ErddapPlatformInfoPanel platform={platform} />
-          <ErddapObservationTable
```
abkfenris, line 25: Cards should end up here or somewhere that is functionally the same location on the site.

src/Features/ERDDAP/Platform/Observations/CurrentConditions/index.tsx
abkfenris: I think the main adaptation would be to get a wave group out of the function as well.

abkfenris (commented):
Ok, chatted it through with Riley and she helped convey Tom's understanding. We want the cards to be back up in the header, where the old 'last updated at' observation table was.

This solves a few things:
- We can see the latest reading via the cards on all the platform pages
- The map doesn't get squeezed too small
- We don't need to duplicate the unit selector elsewhere

To try to clean things up:
- Don't have a filter in cards. They are already combining some of the data types, and it adds a lot of complexity both in code and UX
- The individual cards should be allowed to fill the width, rather than being restricted to 2 wide. This will keep them from ending up too tall, and hopefully end up somewhere near the current 'Last updated at' table.

This is really Rileys first mockup (with the filters from quick photoshopping removed):


>     1. We could replace the table with the cards and keep them above the plots but always visible. This is basically the same behavior as the current site just using cards instead of links to update charts. We'd have more chart control but would lose the tidier look and feel and vertical real estate gained by making the map shorter and only showing station info. We also may need to make the card table more of a landscape mode to better fit the space.
> 
> 
> MDdesigncards

src/Features/ERDDAP/Platform/Observations/LatestObsCards/LatestObsCards.tsx
```diff
@@ -54,21 +52,25 @@ export const ErddapObservationCards: React.FC<Props> = ({
       ) : (
         <div>There is no recent data from {platformName(platform)}</div>
       )}
-      <Row xs={1} lg={2} className="align-items-stretch">
-        {waveTs.length > 0 && <LatestObsCard timeSeries={waveTs} platform={platform} unitSystem={unitSystem} />}
-        {windTs.length > 0 && <LatestObsCard timeSeries={windTs} platform={platform} unitSystem={unitSystem} />}
-        {otherTs.map((ts, index) => {
+      <Row xs={1} lg={3} className="g-1">
```
abkfenris: I think this could be `<Row xs={2} lg={3} xl={4} xxl={5} className="g-1">` to allow sizing more cards per row while keeping it reasonable.

src/Features/ERDDAP/Platform/Observations/LatestObsCards/LatestObsCards.tsx
```diff
@@ -30,13 +29,12 @@ export const ErddapObservationCards: React.FC<Props> = ({
   laterThan,
   children,
 }: Props) => {
-  const { allCurrentConditionsTimeseries } = currentConditionsTimeseries(platform, laterThan)
-  const times = allCurrentConditionsTimeseries.filter((d) => d.time !== null).map((d) => new Date(d.time as string))
+  const { windTimeSeries, waveTimeSeries, allNonGroupedTimeseries } = currentConditionsTimeseries(platform, laterThan)
```
abkfenris:
```ts
const {before, windTimeSeries, waveTimeSeries, timeSeries, after} = currentConditionsTimeseries(platform, laterThan)
```

No idea why it isn't letting me select diffs today.

abkfenris (commented):
Looking good!

> Something I need clarity on -- this implementation does not render all highlighted data from `currentConditionsTimeseries`, only cards relevant to wind, wave, air temp, SST, and air pressure are being shown. Is this what we're looking for?

We want all the highlighted data from `currentConditonsTimeseries`, so roughly `{before, windTimeSeries, waveTimeSeries, timeSeries, after}` and displaying them in the same order.

For example, the Boothbay Tide Gauge has the tide data highlighted, so that should show in a card.

image

If mobile and iPad mini sizes can fit two cards horizontally, I'd prefer that to reduce the height of the block overall.



to

image

and 

image

I think really wide can also go more than 3 wide.

image

to

image

src/Features/ERDDAP/Platform/Observations/LatestObsCards/LatestObsCards.tsx
shindelr: Yeah the diffs are coming out a little funny today. Thanks for bearing with me while I figure out currentConditionsTimeseries! I basically duplicated timeSeries with my allNonGroupedTimeseries. Nuked it and went back to the original implementation

abkfenris (approved):
Looks good! The only thing that I noticed is that a few platforms say that there is no recent data, but then show a card of recent data. I wouldn't show the cards in that case if there isn't a timestamp right there for folks to judge if it is recent enough.

Image

shindelr:
Good catch, I just pushed up a change that should fix the problem. `timeSeries` was coming back full of null objects but normally the table uses `allCurrentConditionsTimeseries` which seems to fill any gaps with `before` and `after`. 

The way everything's wired up, using all the timeseries will cause recurring cards to render and it's easiest to filter out what we don't want back at the source.

Issue #3884, opened 2026-01-12
MD Refresh - Station Detail - Tabs and Large Graphs


youngmorse:
Continuation of Issue #3877 - Station Detail Tabs and Large Graphs

Large Graphs

<img width="1091" height="772" alt="Image" src="https://github.com/user-attachments/assets/c7bf6fee-8ac2-4f9a-aa74-0b7a9a54c546" />

**Back Button**
- Added button to go back to all graphs when clicking on a graph
- Icon: ‘arrow-left’

**Time Lapse Button Group**
- Added time configurations for last 24 hours, last 7 days, and last 30 days
- Unselected
  - Background color: White
  - Stroke: 1px, Black 20
- Selected
  - Background color: Black 5
  - Stroke: 1px, Black 65
- Border radius (ends only): 6px

**Graph Title**
- Title: ‘H2’, Black
- Replaced ‘info’ icon with icon: ‘export-variant’
- Icon color: Black
- Icon button
  - Background color: White
  - Stroke: 1px, Black 20
  - Padding: 8px 10px 8px 10px
  - Border radius: 6px

PR #4099, opened 2026-05-07
Large graph title & export icon

shindelr:
@cgalvarino 
#3884

Title changed to H2, Icon changed and styled. Note the thickness of the icon and the different shape vs. the wireframe. We're using the free icons so this is the weight and shape we have to work with. I experimented with altering the fill/stroke via css but the icon became malformed.

**Graph Title**
Title: ‘H2’, Black
Replaced ‘info’ icon with icon: ‘export-variant’
Icon color: Black
Icon button
Background color: White
Stroke: 1px, Black 20
Padding: 8px 10px 8px 10px
Border radius: 6px

**Goal**
Screenshot 2026-05-07 at 12 31 20 PM

**After**
Screenshot 2026-05-07 at 12 32 17 PM

**Before**
Screenshot 2026-05-07 at 12 31 43 PM

### DIFF
--- a/src/Features/ERDDAP/Platform/Observations/Condition/Info.tsx
+++ b/src/Features/ERDDAP/Platform/Observations/Condition/Info.tsx
@@ -6,6 +6,7 @@ import Popover from "react-bootstrap/Popover"
 import ListGroup from "react-bootstrap/ListGroup"
 
 import { tabledapProtocolUrl } from "Shared/erddap/tabledap"
+import { ExportIcon } from "Shared/icons/iconsMap"
 
 import { PlatformTimeSeries } from "../../../types"
 
@@ -64,11 +65,7 @@ export const Info: React.FC<InfoProps> = ({ timeSeries, id, startDate }: InfoPro
   return (
     <React.Fragment>
       <OverlayTrigger delay={{ show: 250, hide: 4000 }} overlay={renderToolTip} trigger={["click"]}>
-        <FontAwesomeIcon
-          id={`tooltip-${id}-trigger`}
-          icon={faInfoCircle}
-          style={{ fontSize: "1rem", verticalAlign: "middle" }}
-        />
+        <ExportIcon id={`tooltip-${id}-trigger`} className="export-icon-border bg-white p-1" />
       </OverlayTrigger>
     </React.Fragment>
   )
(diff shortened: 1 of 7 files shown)


src/Features/ERDDAP/Platform/Observations/Condition/Info.tsx
```diff
@@ -64,11 +65,7 @@ export const Info: React.FC<InfoProps> = ({ timeSeries, id, startDate }: InfoPro
   return (
     <React.Fragment>
       <OverlayTrigger delay={{ show: 250, hide: 4000 }} overlay={renderToolTip} trigger={["click"]}>
-        <FontAwesomeIcon
-          id={`tooltip-${id}-trigger`}
-          icon={faInfoCircle}
-          style={{ fontSize: "1rem", verticalAlign: "middle" }}
-        />
+        <ExportIcon id={`tooltip-${id}-trigger`} className="export-icon-border bg-white p-1" />
```
abkfenris, line 68: We can probably fix that it doesn't show a pointer by wrapping the icon in a button component, rather than CSS styling.

I think the lack of pointer was something that we didn't notice in the migration between React Bootstrap libraries. React-Bootstrap largely shops their overlay components wrapping Button components https://react-bootstrap.netlify.app/docs/components/overlays/ where as Reactstrap's popovers worked differently and we were controlling the state ourselves https://github.com/shindelr/Neracoos-1-Buoy-App/blob/56b9752dfbc19c4ef1f4079002d7015e5a957313/src/Features/ERDDAP/Platform/Observations/Condition/Info.tsx#L22-L62

src/index.scss
```diff
@@ -38,6 +38,10 @@ $dig-colors: (
   "orange": $orange,
 );
 
+$black-20: rgba(0, 0, 0, 0.2);
+$black-65: rgba(0, 0, 0, 0.65);
+$seaweed-10: rgba(23, 133, 150, 0.1);
+
```
abkfenris: If I'm reading the SASS specs right, I think it's got a function for setting the alpha of an existing color. That way we can have the color variables flow through, rather than having to play 'find the matching tuples'.

https://sass-lang.com/documentation/modules/color/#transparentize

```suggestion
@use 'sass:color';  // somewhere

$black-20: transparentize($black, 0.2);
$black-65: transparentize($black, 0.65);
$seaweed-10: transparentize($seaweed, 0.1);

```

abkfenris (approved):
It works, but I added two notes that may be able to clean things up, including what @cgalvarino noticed with the pointer.

src/index.scss
shindelr: Great find, transparentize is exactly what I needed here

PR #4104, opened 2026-05-13
Map key changed to Recent Data and Out of Date

shindelr:
@cgalvarino 

Quick change to the map key prompted by the Slack conversation.

**Before**
Screenshot 2026-05-13 at 11 17 33 AM

**After**
Screenshot 2026-05-13 at 11 16 34 AM

I did have a playwright change that's not related to the map key, for the life of me I couldn't figure out what was different between dev main and this branch, both of which looked just like prod in the context of what was being tested. Tested main on playwright and the same test failed, so something must've slipped through? Just wanted to highlight this in case you all thought I was pulling a fast one by not mentioning it.


abkfenris (approved):
A01 was recently re-deployed and sometimes some of the data times out during that process. I just gave the backend a kick and everything should be showing, so that Playwright tweak should be unnecessary.

PR #4108, opened 2026-05-18
Large graph quick timeframe selectors

shindelr:
@cgalvarino 
#3884 

Adds a button group to render a large graph of the given observation(s) for a particular span of time. Presets are currently limited to 24 hours, 7 days, and 30 days but can be extended/removed/etc..

Note, styling for specifically for mobile/tablet hasn't been implemented yet. I wanted to pause and ask for guidance since the wireframes don't include that particular information. We were thinking perhaps turning the button group into a dropdown?

 Also looking for guidance on possibly combining the timeframe date picker with this new button group. From Charlton's recent email:
> Robin has been working on the new time controls when in the large graph mode.  The start and end pickers are giving us grief.  The mockup has them visible all the time.  But, as you can imagine, it’s easy for it to get out of sync w/ what time control button may be pressed.  Should the start and end times update to match what button you have mashed (e.g., the 24-hour or 2-weeks button)?  How about this instead:  Add a “Custom” (or something) option.  Only when you hit that button  >would you see the date picker and be able to continue down that rabbit hole.

**Before**
Screenshot 2026-05-18 at 4 40 56 PM

**After**
Screenshot 2026-05-18 at 4 39 57 PM

**Mobile for your reference**
Screenshot 2026-05-18 at 4 42 57 PM

### DIFF
--- a/src/Features/ERDDAP/Platform/Observations/Condition/Info.tsx
+++ b/src/Features/ERDDAP/Platform/Observations/Condition/Info.tsx
@@ -65,7 +65,7 @@ export const Info: React.FC<InfoProps> = ({ timeSeries, id, startDate }: InfoPro
   return (
     <React.Fragment>
       <OverlayTrigger delay={{ show: 250, hide: 4000 }} overlay={renderToolTip} trigger={["click"]}>
-        <ExportIcon id={`tooltip-${id}-trigger`} className="export-icon-border bg-white p-1" />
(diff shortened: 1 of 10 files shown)


abkfenris (commented):
It's still blocking plotting if you go over a year, and then reduce to less than one.

Image

abkfenris (approved):
Looks good!

PR #4115, opened 2026-05-20
Patch to filter out duplicate data in allCurrentConditionsTimeseries.

shindelr:
@cgalvarino

allCurrentConditionsTimeseries.ts was returning duplicate data in the final before and after arrays. To allow for backwards compatibility, I added two arrays -- beforeWithoutGroups and afterWithoutGroups. These just don't include data from the wind and wave groups.

I believe the small charts were crashing because duplicate queries were being sent to backend.

image

### DIFF
--- a/src/Features/ERDDAP/utils/currentConditionsTimeseries.ts
+++ b/src/Features/ERDDAP/utils/currentConditionsTimeseries.ts
@@ -50,15 +50,37 @@ export function currentConditionsTimeseries(platform: PlatformFeature, laterThan
   const timeSeriesWithNull = [waterTemp, airTemp, airPressure, waterLevel, visibility]
   const timeSeries = timeSeriesWithNull.filter((ts) => ts !== null) as PlatformTimeSeries[]
 
-  const nonGroupTimeSeriesWithNull = [...before, waterTemp, airTemp, airPressure, waterLevel, visibility, ...after]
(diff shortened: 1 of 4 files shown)


src/Features/ERDDAP/utils/currentConditionsTimeseries.ts
```diff
@@ -50,15 +50,37 @@ export function currentConditionsTimeseries(platform: PlatformFeature, laterThan
...
+    ...windTimeSeries,
+    ...waveTimeSeries,
+    ...timeSeries,
+    ...afterWithoutGroups,
+  ]
+  const allCurrentConditionsTimeseries = allWithNull.filter((ts) => ts !== null)
+
+  console.log(nonGroupTimeSeries)
+  console.log(waveTimeSeries)
+  console.log(allCurrentConditionsTimeseries)
```
abkfenris: Leftover console logs

abkfenris (approved):
Hmm, duplicate queries to a timeseries should be combined into single requests by react-query rather than cause things to crash, but better to have those explicitly excluded from display.

src/Features/ERDDAP/utils/currentConditionsTimeseries.ts
shindelr: Ripped em out, thanks!

PR #4117, opened 2026-05-20
new regions

cgalvarino:
Your prep made my life much easier!

image

The linter is going to have me for lunch.  Standby.

@shindelr

### DIFF
--- a/src/Features/ERDDAP/Platform/Observations/Table/table.spec.tsx
+++ b/src/Features/ERDDAP/Platform/Observations/Table/table.spec.tsx
@@ -6,35 +6,37 @@ import { render, screen } from "@testing-library/react"
 import { UnitSystem } from "Features/Units/types"
 import { PlatformFeatureWithDatasets } from "../../types"
 
-import { ErddapObservationTable } from "./table"
+import { ErddapObservationCards } from "../LatestObsCards/LatestObsCards"
 
 describe("<ErddapObservationTable>", () => {
   it("Should show selected observations for appropriate platform", () => {
     render(
-      <ErddapObservationTable
+      <ErddapObservationCards
         platform={platform}
         unitSystem={UnitSystem.english}
         unitSelector={<b>Fake unit selector</b>}
       />,
     )
 
-    expect(screen.getAllByRole("link").length).toBe(6)
-    expect(screen.findByText("Last updated at"))
-    expect(screen.findByText("Wind Speed: 3.9 kts"))
+    expect(screen.getAllByRole("link").length).toBe(2)
+    expect(screen.findByText("Last updated"))
+    expect(screen.findByText("Wind"))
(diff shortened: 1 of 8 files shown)


cgalvarino:
Oh dear.  You know what?  New regions means storybook will probably need some TLC.

@shindelr we'll add this to the list offline.

src/Features/ERDDAP/Platform/Observations/Table/table.spec.tsx
abkfenris, line 1: Is this an unrelated change?

src/Shared/constants/regions.ts
abkfenris, line 1: I can't remember why this is a separate file, but we can probably consolidate it into `src/Shared/regions.ts`, especially if we're going to have a menu and a on-page lists of regions. It also looks like both define `regionList`. 

It would also make it a bit cleaner to test to make sure the regions included in the lists are the same.

abkfenris (commented):
Now I think we really want to have an separate menu and page list of regions.

Image

I'd start by implementing a `regionPageList` or similar (and probably rename `regionList` to `regionMenuList`) and starting with the reverse order, and then I can sit with Tom and Riley and tweak the order later.


@tshyka @youngmorse do we need a region specific error page to catch folks who've navigated to one that no longer exists? Or the map enough?

Image

Maybe the move is to add a list of regions with links to the current null fallback?

https://github.com/gulfofmaine/Neracoos-1-Buoy-App/blob/5492a4ecc755849be9df1e2e7599c42fd8b709e7/app/(platformMap)/%40sidebar/region/%5BregionId%5D/page.tsx#L16-L19

```tsx
  if (typeof region === "undefined") {
    return (
      <>
        <h2>Regions</h2>
        <ul>
        {regionList.map(r => (
          <li key={r.slug}><Link href={`/region/${r.slug}`}>{r.name}</Link></li>
        ))}
        </ul>
      </>
    )
  }
```

Image

src/Shared/constants/regions.ts
youngmorse: We probably didn't notice the lack of a landing page for regions because the first option is the Gulf of Maine group which contains 80% of the sensors. I think if we do want to have a top level page that lists the region options, your mockup makes sense. Are you thinking the top level menu option or adding an 'All platforms' on the list? On the [Water Level dashboard](https://mariners.neracoos.org/water-level) the landing page has all the regions as cards with links to sensors within them but we may have too many stations on MD to list them all out out on one page. 

When I was getting the bbox coordinates, I had them all activated at once and I think there was enough overlap that all the platforms are included in at least one group. 

And we need to be consistent with the name of this feature (?) - Regions/Stations/Platforms. I thought we resolved this with Dig. I will look at figma and update.

abkfenris (approved):

PR #4119, opened 2026-05-21
Card link to wind combo plot

cgalvarino:
@shindelr plz make sure this has the desired effect.


shindelr (approved):

shindelr:
Those tests were failing this morning but I know that on #4126 they're passing just fine but with the added mock-time code. I reset back to your original commit @cgalvarino just now. We'll see if it passes otherwise it may need the same treatment as the other PR

PR #4159, opened 2026-06-01
Stacked charts on wave page

shindelr:
@cgalvarino 

When a user clicks on the grouped latest obs wave card, they're taken to `observations/waves` which shows all wave related conditions charts from the station.

Didn't implement this sort of grouping for when someone clicks on the small chart cards on a station. Can definitely add that in though!

Tough to get good screenshots of this behavior.
Screenshot 2026-06-01 at 9 20 08 AM

Screenshot 2026-06-01 at 9 19 39 AM

### DIFF
--- a/src/Features/ERDDAP/Platform/Observations/Condition/index.tsx
+++ b/src/Features/ERDDAP/Platform/Observations/Condition/index.tsx
@@ -18,6 +18,7 @@ import { UseDataset } from "../../../hooks"
 import { PlatformFeature, PlatformTimeSeries } from "../../../types"
 import { Info } from "./Info"
 import { getStartFunction, Timeframes } from "Features/ERDDAP/TimeframeButtonGroup/timeframes"
+import { read } from "node:fs"
 
 interface Props {
   /** Platform to display */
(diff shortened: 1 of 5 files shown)


src/Features/ERDDAP/Platform/Observations/Condition/index.tsx
```diff
@@ -18,6 +18,7 @@ import { UseDataset } from "../../../hooks"
 import { PlatformFeature, PlatformTimeSeries } from "../../../types"
 import { Info } from "./Info"
 import { getStartFunction, Timeframes } from "Features/ERDDAP/TimeframeButtonGroup/timeframes"
+import { read } from "node:fs"
```
abkfenris: I don't think we should be importing from `"node"`

src/Features/ERDDAP/Platform/Observations/Condition/index.tsx
```diff
@@ -49,9 +50,12 @@ export const ErddapObservedCondition: React.FunctionComponent<Props> = ({ platfo
...
 
-  const timeSeries: PlatformTimeSeries[] = platform.properties.readings.filter(
-    (reading) => reading.data_type.standard_name === standardName,
-  )
+  // Special case to handle grouped wave latest obs card
+  const timeSeries: PlatformTimeSeries[] =
+    standardName === "waves"
+      ? platform.properties.readings.filter((reading) => reading.data_type.long_name.match("Wave"))
+      : platform.properties.readings.filter((reading) => reading.data_type.standard_name === standardName)
+
```
abkfenris: This doesn't feel like the right place for having this logic.

The special casing for the wind chart happens at the routing level by having a page at https://github.com/gulfofmaine/Neracoos-1-Buoy-App/blob/main/app/(platformMap)/%40bottom/platform/%5BplatformId%5D/observations/wind/page.tsx

I'd prefer to mirror that structure and have a page at `app/(platformMap)/@bottom/platform/[platformId]/observations/waves/page.tsx` to show the stacked charts.

abkfenris (changes requested):

src/Features/ERDDAP/Platform/Observations/Condition/index.tsx
shindelr: Yep, must've been an intellisense mishap

src/Features/ERDDAP/Platform/Observations/Condition/index.tsx
shindelr: Completely makes sense to me! I'll wire up a page.tsx under `observations/waves`.

We'll still need this logic for filtering the readings to catch all the waves though, correct? To follow the same pattern as wind we'd create a new dir/index combo like `WaveCondition/index.tsx` and specify that filter there?

abkfenris (approved):
I'm going to get this merged before Tom presents the updates today.

